PR #60, opened 2019-02-25
Demo mode 2.8x

jeromecombes:
Add demo mode
Prevents users from changing the configuration and passwords when demo mode is enabled
To enable demo mode : add $config['demo'] = true; to config.php
PR for branch 2.8x


jeromecombes:
Replaced with PR #73 for master / release 19.04

PR #112, opened 2019-10-25
Fix encoding errors on tables absences, config and personnel

jeromecombes:


jeromecombes:
Replaced with PR #130

PR #125, opened 2019-11-19
checkbox way of getting they're checked or not

flartet:
Hello bonjour,
I corrected the "Mail test" button behaviour. smtpauth was always '1' because of JS
I think you did the same mistake for all the checkboxes. checkbox.val() always returns the value.
Also the line was duplicated but there was no problem about it.


jeromecombes:
Hello,
Thanks for your help.
It's already fixed by PR #120 and #121 merged on branches master and 19.10.xx
All the best,
Jérôme

PR #128, opened 2019-11-27
POC: add i18n capabilities

alexarn:
Add i18n capabilities and make /absences/info translatable.

Test plan:
  - Go to /absences/info
  - Check the language corresponds to your brower language (only fr and en supported),
  - force another language by adding a locale parameter in url: /absences/info?locale=en or  /absences/info?locale=fr
  - Refresh the page and check the language is kept in session,
  - Try a unsupported langue: /absences/info?locale=es
  - Check the language move to default: en

To do:
  - Generate a messages file with all string,
  - Translate all the string!

This works only for parts rendered by a controller.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -26,9 +26,9 @@ services:
         resource: '../src/Controller'
         tags: ['controller.service_arguments']
 
-    App\Listener\ControllerAuthorizationListener:
+    App\Listener\KernelRequestListener:
         tags:
-            - { name: kernel.event_listener, event: kernel.request }
+            - { name: kernel.event_listener, event: kernel.request, priority: 20 }
 
     # add more service definitions when explicit configuration is needed
     # please note that last definitions always *replace* previous ones
--- a/src/Controller/AbsenceInfoController.php
+++ b/src/Controller/AbsenceInfoController.php
@@ -17,6 +17,7 @@ class AbsenceInfoController extends BaseController
      */
     public function index(Request $request, Session $session)
     {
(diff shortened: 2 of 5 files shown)


veggiematts:
Tested ok according to test plan.

In order to be ready for all future possible translations and browser settings, it would be nice to support country as well as language (ie: fr_FR is not the same as fr_CA or fr_BE).

Question: why use yaml files instead of gettext files ? Gettext is a standard for translations, with many benefits, like having translation softwares that would allow non-developers to translate, or the support of plural forms, for instance.

alexarn:
> Tested ok according to test plan.
> 
> In order to be ready for all future possible translations and browser settings, it would be nice to support country as well as language (ie: fr_FR is not the same as fr_CA or fr_BE).
> 
> Question: why use yaml files instead of gettext files ? Gettext is a standard for translations, with many benefits, like having translation softwares that would allow non-developers to translate, or the support of plural forms, for instance.



> In order to be ready for all future possible translations and browser settings, it would be nice to support country as well as language (ie: fr_FR is not the same as fr_CA or fr_BE).

I agree !

veggiematts:
Apparently plural forms are supported in symfony, regardless of the catalog file format (yaml for instance):
https://symfony.com/doc/3.4/components/translation/usage.html#component-translation-pluralization

There also seems to be a gettext file loader:
https://symfony.com/doc/current/components/translation.html :

> PoFileLoader - to load catalogs from gettext files.

PR #137, opened 2020-02-06
MT 27553: rename recup_samedi to comp_time (19.10.xx and 19.04.xx)

alexarn:


alexarn:
Merged into 19.10.xx and 19.04.xx

PR #138, opened 2020-02-12
MT 28331: saving services or status doesn't take into account the appropriate values

alexarn:


alexarn:
Fix applied on 2.7x, 19.04.xx, 19.10.xx and master

PR #143, opened 2020-03-23
MT 28713: Correct uses of App\Model\Agent in recovery system

alexarn:
holiday recovery system. Also, remove a php warning.


alexarn:
Pull request merged manually

PR #145, opened 2020-03-31
Prevent non authorized users to see holidays

alexarn:


alexarn:
Replaced with PR #147

PR #152, opened 2020-04-08
Fix wrong SQL queries on the departure date in ajax.menudiv.php

jeromecombes:
Agents remained visible in the "unavailable agents" menu despite a departure date exceeded, due to wrong SQL queries (missing quotes).


alexarn:
Merged on matser

PR #166, opened 2020-05-13
Change order of absences settings

jeromecombes:
Also add Absences-DelaiSuppressionDocuments in db_data.php


alexarn:
Merged on master

PR #184, opened 2020-06-10
Enable web profiler toolbar on dev environnement. Add APP_DEBUG=0 in …

jeromecombes:
Enable web profiler toolbar on dev environnement.
Simply add APP_DEBUG=0 in .env.local to disable it


alexarn:
Merged on master

PR #186, opened 2020-06-10
Update PHPMailer

jeromecombes:
PHPMailer 6.1.6
This is a security release, with some other minor changes. For full details, refer to the advisory.

SECURITY Fix insufficient output escaping bug in file attachment names. CVE-2020-13625. Reported by Elar Lang of Clarified Security.
Correct Armenian ISO language code from am to hy, add mapping for fallback
Use correct timeout property in debug output


alexarn:
Merged on master

PR #188, opened 2020-07-01
Lock planning table config

alexarn:
Currently, users are able to edit planning table configuration that are used in some plannings. So they can:
  - Change sub-tables number,
  - Edit time ranges,
  - change or remove jobs and disable related time range.

That can lead to corrupted data if some agents are already placed on these modified cells/jobs.

To reproduce the issue:
 - Create a planning table (/index.php?page=planning/postes_cfg/modif.php)
 - Use this table for a planning,
 - Add an agent in the planning (where you want)
 - Back to table editing and change the time range you placed the agent on.
 - Back on the planning => agent disappeared. (but still in database),
 - Revert your change on time range => agent is back!
 - Do the same with jobs (change or remove) and disable time ranges in "Lignes" tab
   Each time, agent disappears but is still in database.

Test plan for submitted patches:
- Create a planning table with sub-tables, jobs and time ranges inside etc ...
- You can change/add/edit all you want inside,
- use this table for a planning (don't put agents inside),
- check that you can edit all you want except the site (disabled) in "infos générales" tab,
- Add an agent in the planning,
- check that you can't change the sub table number in "infos générales" tab,
- check that the time range you put the agent on is disabled in "Horaires" tab,
- check that the job select you put the agent on is disabled in "Lignes" tab,
- check that you can't enable/disable the related time range in "Lignes" tab.


alexarn:
Just realized ther's a tiny mess in my test plan:
 - Use a table for a planning,
 - check that you can't delete the table.

Not sure of that. Should we lock the table deletion only if there is at least one agent inside ?
Any thoughts on this ?

alexarn:
Replaced with PR #489 
So close

PR #189, opened 2020-07-10
update install.sh : remove composer checksum and handle remote hosts

jeromecombes:
update install.sh : 
1./ remove composer checksum
-- the checksum changes with new releases of composer. Keeping an old checksum makes the download fail.
2./ handle remote hosts
-- install.sh was not able to create DB on remote hosts.

I also changed the process order : 
1./ Create the DB, stop if errors
2./ Create .env.local
3./ Download and install the theme
4./ Download and install composer
5./ Execute composer install
And add this information : "run chmod -R 777 var"

update public/setup/maj.php :
-- fix a typo : remove Bold

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -143,16 +143,32 @@ if [[ $updatecomposer = '' ]]; then
 fi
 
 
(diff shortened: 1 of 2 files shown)


alexarn:
Merged on master

PR #191, opened 2020-07-16
MT28388: Cron script to purge the log table

veggiematts:
The delay parameter is mandatory. See https://dev.mysql.com/doc/refman/en/expressions.html#temporal-intervals.

Usage:   php bin/console PlanningBiblio:PurgeLogTable "<DELAY>"
Example: php bin/console PlanningBiblio:PurgeLogTable "12 MONTH"

Test plan:
 - execute the command with a given <DELAY>
 - check that logs older than <DELAY> are removed from the log table.

### DIFF
--- a/src/Cron/PurgeLogTable.php
+++ b/src/Cron/PurgeLogTable.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace App\Cron;
+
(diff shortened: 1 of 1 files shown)


alexarn:
Merged on master

PR #192, opened 2020-07-30
MT27842: Having "Full day" checked by default in absences is configurable.

veggiematts:
Test plan:
    
     1) apply the patch and the database update
     2) Check that "Full day" is unchecked when adding an absence
     3) Set "Absences-journeeEntiere" configuration option to true
     4) Check that "Full day" is checked when adding an absence


alexarn:
Merged on master

PR #196, opened 2020-08-17
Fix removal of reccuring absences (master)

jeromecombes:
Fix removal of reccuring absences
- Fix wrong links in file public/absences/js/modif.js
- Add hidden input "baseURL" in templates/base.html.twig to get the base URL into JS files


alexarn:
Merged on master

PR #197, opened 2020-08-17
Fix removal of reccuring absences (20.04.xx)

jeromecombes:
Fix removal of reccuring absences **for branch 20.04.xx**

Fix wrong links in file public/absences/js/modif.js
Add hidden input "baseURL" in templates/base.html.twig to get the base URL into JS files


alexarn:

PR #199, opened 2020-08-20
Set default value for start and end hours when saving an absence (master)

jeromecombes:
Set default value for start and end hours when saving an absence


alexarn:
Merged on master

PR #200, opened 2020-08-20
Set default value for start and end hours when saving an absence (20.04.xx)

jeromecombes:
Set default value for start and end hours when saving an absence


alexarn:

PR #202, opened 2020-08-20
Fix admin info and absences info forms (master)

jeromecombes:
Fix admin info and absences info forms (master)
réf : MT 29802


alexarn:
merged on master

PR #203, opened 2020-08-20
Fix admin info and absences info forms (20.04.xx)

jeromecombes:
Fix admin info and absences info forms (20.04.xx)
Réf. : MT29802


alexarn:

PR #213, opened 2020-09-14
MT 30038: Use fine tuning of hours in holidays forms for master

jeromecombes:


alexarn:
"hour_precision" template parameter can be removed in AbsenceController::edit, right ? This is done in the same PR for 20.04

jeromecombes:
> "hour_precision" template parameter can be removed in AbsenceController::edit, right ? This is done in the same PR for 20.04

Right !

alexarn:
Merged on master

PR #223, opened 2020-09-25
Update README, required : add php-xml, remove mysql-client/server

jeromecombes:
Update README, required : add php-xml, remove mysql-client/server


alexarn:
Merged on master for release 20.10

PR #226, opened 2020-09-30
symfonize stats time

sillydebs:


alexarn:
Agent' status have html entities. May be a raw filter could fix that.
For the rest, seems ok.

alexarn (changes requested):

alexarn:
Replaced by #356

PR #230, opened 2020-10-06
MT 30150: SSL errors can occur during MS Graph imports.

veggiematts:
This patch adds 4 retries when an error occurs during an Unirest request.


alexarn (approved):

alexarn:
Merged on 20.10.xx

PR #250, opened 2020-10-20
MT 30359 - Keep absence's other reason when there is recurrence

alexarn:


alexarn:
Next step would be to change the absence table (list) to show the other reason instead of "Autre".
As there is currently a work on this page, i think this should be done in PR 207 or after the merge.

Any opinions ?

alexarn:
Merged on master and 20.10.xx

PR #252, opened 2020-10-21
Ability to change login in agent's form

alexarn:


veggiematts:
Tested ok, but some filtering is needed ( <script> alert('coincoin'); </script> is a nice login to try :))

PR #257, opened 2020-10-28
Update DataTable libraries and remove flash buttons

jeromecombes:


alexarn:
Merged on master and 20.10.xx

PR #265, opened 2020-11-06
Symfonize planning/postes_cfg/lignes.php

sillydebs:


src/Controller/FrameworkController.php
```diff
@@ -100,4 +100,88 @@ public function index (Request $request, Session $session){
...
+    public function addLine (Request $request, Session $session){
+        $CSRFToken = $GLOBALS['CSRFSession'];
+        $id = null;
+        $nom = null;
+
+        $this->templateParams(
+            array(
+                "CSRFToken"    => $CSRFToken,
+                "id"           => $id,
+                "nom"          => $nom
```
alexarn, line 116: These 2 variables (id and nom) are useless here. When adding a new line, we never have id or nom

alexarn (changes requested):
Encoding issue:
If you edit a line previously added (with legacy code) with accent, there an encoding issue.
I.e: "Rez de chaussée"

src/Controller/FrameworkController.php
sillydebs: They are required as parameters for the template, thus I have to declare them.

alexarn:
Merged on master

PR #284, opened 2020-12-04
fix bug on position edit

sillydebs:
I noticed on master the following bug : accessing to a position's editing page returns an notice to get the group and the floor to display.


jeromecombes (approved):
à intégrer sur master et 20.10.xx

alexarn:
Merged on master and 20.10.xx

PR #297, opened 2020-12-14
Symfonize agents deletion

sillydebs:


alexarn:
Replaced with PR #445

PR #298, opened 2020-12-15
Introduce YAML file to access management

sillydebs:


alexarn:
This works fine for me. Thx Sarah !

Now, i'd like to go further:
With this feature, with still need paths to be defined in database (table acces) with minimal right to pass the control in init/init.php and then to be re-checked in the authorization listener.

We should read the access.yaml in the init, and skip the database control if the requested route is defined in the yaml.
Make sure we read the yaml only once.

Suggestion: rename access.yaml to permissions.yaml

@Jérôme, any thoughts ?

alexarn:
Now we have an unexpected but understanding behavior:
  - init.php doesn't worried about mapping between paths and routes names (this is the job of the kernel)
  - So, $request->attributes->get('_route') in the init return nothing.

Solution: move all the code related to rights checking in the listener

This include:
  - getting accesses, 
  - checking DB authorization,
  - checking YAML authorization,
  - checking anonymous auth mode,
  - handle specific case for ajax routes

src/Listener/ControllerAuthorizationListener.php
```diff
@@ -15,25 +20,48 @@ class ControllerAuthorizationListener
...
+
+        // Droits necessair<es pour consulter la page en cours
+        $accesses = $this->entityManager->getRepository(Access::class)->findBy(array('page' => $page));
+        $logged_in = $this->entityManager->find(Agent::class, $_SESSION['login_id']);
+
+        $authorized = true;
         $route = $event->getRequest()->attributes->get('_route');
 
-        if (!$this->canAccess($route)) {
+        if(!in_array($route, $this->permissions)){
```
alexarn, line 52: This condition is always true because in_array search the value in $this->permissions not the key. use empty($this->permissions[$route]).

src/Listener/ControllerAuthorizationListener.php
```diff
@@ -15,25 +20,48 @@ class ControllerAuthorizationListener
...
+            $authorized = $logged_in ? $logged_in->can_access($accesses) : false;
```
alexarn, line 53: I'd like we return an access denied right here if $logged_in->can_access is false instead of checking $authorized variable later

src/Listener/ControllerAuthorizationListener.php
```diff
@@ -15,25 +20,48 @@ class ControllerAuthorizationListener
...
+
+        if ($_SESSION['oups']["Auth-Mode"] == 'Anonyme' ) {
+            foreach ($accesses as $access) {
+                if ($access->groupe_id() == '99') {
+                    $authorized = true;
+                }
+            }
+        }
+
+        if (!($this->canAccess($route) and $authorized == 'true')) {
```
alexarn, line 64: $authorized == 'true' and $authorized == true are differents.
Here $authorized will *never* equal to the string 'true'.

Anyway, we don't need this check if we early return in the condition line 52.

alexarn (changes requested):
If you try to change agents password => Error. This is because ajax routes are not handled in the listener. See line 31 in public/index_symfony.php

src/Listener/ControllerAuthorizationListener.php
```diff
@@ -15,25 +20,48 @@ class ControllerAuthorizationListener
...
+        if(!in_array($route, $this->permissions)){
+            $authorized = $logged_in ? $logged_in->can_access($accesses) : false;
+        }
+
+        if ($_SESSION['oups']["Auth-Mode"] == 'Anonyme' ) {
+            foreach ($accesses as $access) {
+                if ($access->groupe_id() == '99') {
+                    $authorized = true;
+                }
+            }
```
alexarn, line 61: Early return an access-denied here please

alexarn (changes requested):
The variable $authorized and the code modifying $page are no longer needed in the init

alexarn:
Merged on master

PR #301, opened 2020-12-16
MT 31000 - Retrieve correct public holiday name in calendar (master and 20.10.xx)

alexarn:


jeromecombes (approved):

alexarn:
Merged on master and 20.10.xx

PR #302, opened 2020-12-17
MT 30866: Replace heredoc string by double quotes string at end of weekly planning

alexarn:


alexarn:
Sometimes, i don't know why, the heredoc syntax make some troubles in the html output.
Exemple: the planning notes in weekly planning page is writen in a PHP heredoc. So the footer is skipped.

jeromecombes (approved):

alexarn:
Merged on master, 20.10.xx and 20.04.xx

PR #303, opened 2020-12-17
MT 31000 - Take in account empty site in agent's times in holiday validation workflow

alexarn:
Retrieving site number from agents times could bring an empty value. This can happen when validating holiday on a period whose at least one day has no site set (i.e a saturday: ['', '', '', '']).

According to PHP version or configuration, <?php 400 + '' ?> can just warn or ... crash.

This path fix that

### DIFF
--- a/public/conges/class.conges.php
+++ b/public/conges/class.conges.php
@@ -862,10 +862,9 @@ class conges
                 // Récupération du numéro du site concerné par la date courante
                 $offset=$jour-1+($semaine*7)-7;
                 if (array_key_exists($offset, $temps)) {
-                    if (array_key_exists(4, $temps[$offset])) {
+                    $site = 1;
+                    if (!empty($temps[$offset][4])) {
                         $site=$temps[$offset][4];
-                    } else {
-                        $site=1;
                     }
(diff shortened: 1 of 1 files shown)


public/conges/class.conges.php
```diff
@@ -862,10 +862,9 @@ public function getResponsables($debut=null, $fin=null, $perso_id)
                 // Récupération du numéro du site concerné par la date courante
                 $offset=$jour-1+($semaine*7)-7;
                 if (array_key_exists($offset, $temps)) {
-                    if (array_key_exists(4, $temps[$offset])) {
+                    $site = 1;
+                    if (array_key_exists(4, $temps[$offset]) && $temps[$offset][4]) {
                         $site=$temps[$offset][4];
```
jeromecombes, line 867: Pour le contrôle, je conseille de remplacer ceci :
<pre>if (array_key_exists(4, $temps[$offset]) && $temps[$offset][4]) {</pre>
par ceci
<pre>if (!empty($temps[$offset][4])) {</pre>

jeromecombes (changes requested):

jeromecombes (approved):

alexarn:
Merged on master, 20.10.xx and 20.04.xx

PR #304, opened 2020-12-17
Fix holiday controls on halfdays for 20.10.xx

jeromecombes:


alexarn:
Patched included in PR 306

PR #311, opened 2020-12-23
symfonise notifications' save

sillydebs:


alexarn (changes requested):
You add changes from an other pull request (298) and that have been merged on master already. Why ?
Also, provide a config/permissions.yaml with only the newly added routes

2nd option:
rebase your branch on master

config/permissions.yaml
```diff
@@ -13,3 +13,7 @@
     - 100
 "ajax.holidaycredit":
     - 100
+"notifications.index":
```
alexarn, line 16: Good thing to provide notifications.index. So you can remove the entry in access with page = notification

src/Controller/NotificationController.php
```diff
@@ -9,6 +9,7 @@
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 
+require_once(__DIR__ . '/../../public/include/config.php');
```
alexarn, line 12: Why do you need config.php here ?

alexarn (changes requested):

src/Controller/NotificationController.php
sillydebs: I cannot remember, sorry.

config/permissions.yaml
sillydebs: Thank you. db_data.php and atomicupdate are now updated.

config/permissions.yaml
```diff
@@ -13,3 +13,7 @@
...
+    - 21
+"notifications.save":
+    - 21
```
alexarn, line 19: Routes names are wrong in the permissions.yaml.
In the controller, the are called notification.xxxx
So access denied

alexarn (changes requested):
Error when saving notifications => Access denied

alexarn:
Merged on master

PR #314, opened 2021-01-08
Handle empty groups in framework group page

alexarn:


jeromecombes (approved):

alexarn:
Integrated with symfonization of framework froups

PR #325, opened 2021-01-26
Fix/Statistics : Initialize information about agents (20.10.xx)

jeromecombes:
Fix/Statistics : Initialize information about agents when they are always absent

PR for 20.10.xx, also on #324 

To reproduce :
 - Put a few agents on a planning
 - Add an absence for one of them for the whole day
 - See statistics by agent for this day (focus on this day only to be sure that the absence cover the whole period)
--> The agent who is absent is not correctly initialized. He takes the name of the previous agent, or no name if he is on the first position.

### DIFF
--- a/public/statistiques/agents.php
+++ b/public/statistiques/agents.php
@@ -243,13 +243,6 @@ if (!empty($agents)) {
                             $exists_JF=true;
                         }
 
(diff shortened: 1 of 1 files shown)


alexarn:
Merged on 20.10.xx

PR #332, opened 2021-01-29
Handle quotes in skills batch modification (20.10.xx)

alexarn:
Test plan:

  - Create a skill with a quote in name (your's),
  - Select some agent in list (personnel/index.php) and click on edit selected,
  - Try to add skills => fail


alexarn:
Merged on 20.10.xx

PR #337, opened 2021-02-04
MT29618: Add hours to days custom conversion

veggiematts:
An hours to days array can be defined in custom_options.php:
    
    $config['conges-hours-per-day'] = array(
        '175.50' => 6.68,
        '0' => 6,
    );
    
    This array is meant to be understood this way:
     - from 0 (included) to 175.50 (excluded) hours of holidays a year, a day is 6 hours
     - from 175.50 (included) hours of holidays a year, a day is 6.68 hours
    
    Please note that quotes are mandatory when using decimal numbers for hours in the
    array (because php treats floats as integers when used as array keys).
    
    Hours to days conversions will only be displayed if the array is set.

When enabled, hours to days conversions can be seen in the following pages:
  - Holidays-> Recovers (public/conges/recuperations.php)
  - Administration -> agents -> Holidays (templates/agents/holidays_hours.html.twig)
  - Holidays -> list (templates/conges/index.html.twig)
  - Holidays -> add (templates/conges/add.html.twig)
  - Holidays -> edit (templates/conges/edit.html.twig)


jeromecombes (approved):

alexarn:
First 9 patches squashed. All is merged on master. Thanks !

PR #338, opened 2021-02-05
MT29617: Multiple agents for holidays

veggiematts:
If the user doesn't have any of the following rights:
        "Gestion des congés, validation niveau 1"
        "Gestion des congés, validation niveau 2"
    => The selected user is the logged-in user, and this is not editable.
    
    If the user has at least one of the following rights:
        "Gestion des congés, validation niveau 1"
        "Gestion des congés, validation niveau 2"
    => The user can choose one (and only one) agent
    
    If the user has at least one of the following rights:
        "Gestion des congés, validation niveau 1"
        "Gestion des congés, validation niveau 2"
    and the configuration option "Conges-Recuperations" is set to "Dissocier":
    => The user can choose one or more agents (but not all agents at the same time)
    
    If the user has at least one of the following rights:
        "Gestion des congés, validation niveau 1"
        "Gestion des congés, validation niveau 2"
    and the configuration option "Conges-Recuperations" is set to "Dissocier"
    and the configuration option "Conges-tous" is enabled:
    => The user can choose one or more agents, or all agents at the same time.
    
    This patch also adds a per site filter. All sites are checked by default.


jeromecombes:
Déjà intégré

PR #341, opened 2021-02-05
Absences multi-sites: show only managed agents in list and in add form

jeromecombes:


jeromecombes:
This commit is already on 20.10.xx
--> PR for master

alexarn:
Merged on master

PR #344, opened 2021-02-10
symfonize groupes' deletion in framework's page

sillydebs:


alexarn:
Work already done in PR #321. Close

PR #345, opened 2021-02-10
symfonize lines' deletion for frameworks

sillydebs:


alexarn:
Work already done in PR #320

PR #349, opened 2021-02-11
Statedweeks : allow modification of hours on main columns

jeromecombes:
Statedweeks : allow modification of hours on main columns


alexarn:
Merged on BSG_20.10.xx

PR #354, opened 2021-02-12
Restore right for self-removing in absence edition (master, 20.10.xx, 20.04.xx)

alexarn:


alexarn:
Merged on master and 20.10.xx

PR #364, opened 2021-02-13
202102 softness

jeromecombes:
Work in progress ...

Set softness theme as theme.
This PR replace #315 

TODO : 
- optimize theme for tables with custom class on plannings
- prevent inheritance on former themes
- clean default.css


jeromecombes:
Replaced with #365

PR #365, opened 2021-02-14
CSS: Softness theme and Font-awesome integration

jeromecombes:
Integration of new theme Softness and font-awesone CSS

Replace #364 and #315 

This new PR :
- loads font-awesone icons on demand, with a file named loader.js in the theme folder
- keeps compatibility with former themes and former icons
- add a drop-down menu to help selection of theme on admin / config page.

Thanks to @sillydebs for her work on this project


alexarn:
In absence form. Icon for removing agent is not visible
In the modal window for managing absences reasons, the baskets for removing reasons are shifted
  => Same in agents form for statuses ans services 

Plannings: In time header, text is yellow on white backgroud.

Framework:
  - In the time tab, select are not aligned
  - In the lines tab, some icon are not visible due to the same background
  - In the lines tab, disabled time slot (grey) haven no border

It would be nice if the icons (edit, clone, delete etc...) could change color on focus. IIRC it was the case before

PR #366, opened 2021-02-15
Bigger planning week title

fridobox:
Planning week title (ie 'Planning du Mardi 9 février 20210') font size is only 10.5pt big.
In comparaison, '.td_postes' has font size 12pt.
This is the most important information of the page, should be at least 12pt big.


alexarn:
Obsolete since we made the new theme

PR #367, opened 2021-02-15
Don't show right arrow on locked plannings or for agents without authorization

alexarn:
Test plan:
  - Create a planning,
  - Place some agents on it,
  - lock the planning,
  => right arrow is still available and clickable

### DIFF
--- a/public/planning/poste/js/planning.js
+++ b/public/planning/poste/js/planning.js
@@ -78,6 +78,11 @@ $(document).ready(function(){
 
   $('td.menuTrigger').hover(
     function() {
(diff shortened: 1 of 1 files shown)


veggiematts:
I can't find a permission case where an agent can display an unlocked planning but cannot modify it: either he can display and modify it, or he can't display it at all.

So this patch works: it prevents the use of the arrow on a locked planning, but I wonder if this check is really necessary:
`$("#planning-data").attr("data-autorisation") != 1`

alexarn:
Merged on master and 20.10.xx

PR #368, opened 2021-02-15
Fix datatable with fixed columns

alexarn:


alexarn:
Merged on master and 20.10.xx

PR #370, opened 2021-02-18
MT31368 - add a param to define the switching time between half-day a…

jeromecombes:
MT31368 - add a param to define the switching time between half-day and full-day

the default time for switching from half-day to full-day is 4 hours (14400 seconds)
It can now be changed by adding $config['holiday_fullday_switching_time'] = X; in custom_options.php


alexarn:
Merged on master (DBRev 20.11.00.009) and 20.10.xx (DBRev 20.10.04.001)

PR #374, opened 2021-02-23
MT29619: The user can select the validation level when adding an holi…

jeromecombes:
…day.

This patch introduces a new configuration preference, enabled by default:
Conges-validation

When enabled, the validation dropdown list is shown.
When disabled, the holiday is automatically validated.


jeromecombes:
just a test

PR #380, opened 2021-02-26
Statistics: remove former folder templates/statistiques

jeromecombes:
- remove former folder templates/statistiques
- fix wrong links on attendeesmissing template
- add links on index template


alexarn:
Merged on master.
File templates/statistics/index.html.twig doesn't exist on 20.10.xx

PR #383, opened 2021-03-04
Hamac import: Optimize log

jeromecombes:
Hamac import: Optimize log
- new param Hamac-debug, default = false. Can be changed on WebUI : administration / configuration / Hamac


alexarn:
Merged on master and 20.10.xx

PR #388, opened 2021-03-06
Fix wrong reset URL on statistics/attendeesmissing

jeromecombes:


alexarn:
Merged on master and 20.10.xx

PR #396, opened 2021-03-19
Mark leaved agents as crossed off when importing models

alexarn:
Test plan:
 - mark an agent left at a later date (departure),
 - add this agent on a planning before the departure,
 - create a model with this planning
 - import this model on a planning at a date after the departure 
=> agent is placed in the planning and marked as present.
 - apply this patch,
 - repeat the test plan
=> agent is marked as unavailable.


alexarn:
Fixed in #404. Close this one

PR #405, opened 2021-03-30
Allows editing own absences if they are not validated or Absences-validation is false

alexarn:


jeromecombes (approved):
Tested and approved

alexarn:
Merged on master and 20.10.xx

PR #407, opened 2021-04-13
Position's groups and floors: restore deleting and prevent deleting used

alexarn:


alexarn:
Test plan:

Edit a position (/position/<id>)
  - add a floor/group
  -   try to delete it
  => Nothing happen (note the JS error)
  - add a group
  - use it in the position
  - edit groups list
  => still deletable (should not)
  - apply this pull request and test again

veggiematts:
Test results:
 - I could delete a floor or group just fine on master, and I also can with this patch (so you're fixing something that wasn't broken for me).
 - I cannot delete a used group anymore with this patch => OK

alexarn:
> * I could delete a floor or group just fine on master, and I also can with this patch (so you're fixing something that wasn't broken for me).

Can you try this:
  - edit a position,
  - create a new group or floor,
  - save the modal,
  - save the position,
  - edit the position again,
  - try to delete the new created group or floor

veggiematts:
Ok, after saving, when editing the position, and trying to delete a group or floor, on master:

> Uncaught SyntaxError: Invalid or unexpected token


with this patch: deletion works.

I'm OK with merging this PR.

alexarn:
merged on master and 20.10.xx

PR #409, opened 2021-04-15
Use route names according to PLB naming convention.

alexarn:
Move saving code to a dedicated route

Restore rights for /absences/document*


jeromecombes (approved):

alexarn:
Merged on master

PR #414, opened 2021-04-22
Move planning absences in its own twig template

alexarn:


alexarn:
Changes made in planning symfonisation

PR #416, opened 2021-04-23
Use session flash bag for notice and error messages in framework delete and restore

alexarn:
Use session flash bag for notice and error messages in framework delete and restore


alexarn:
Merged on master and 21.04.xx

PR #419, opened 2021-04-23
Load an up-to-date SQL file

jeromecombes:
For master
This PR :
- Loads an up-to-date SQL file, which avoids loading SQL queries from the maj.php file for new installation


alexarn:
Merged on master

PR #420, opened 2021-04-26
Use common function for getting managed agents in holiday and comptime controllers

alexarn:
Test plan:

- Enable multi-sites,
- create or use an admin  account with right 401 and 601 (but not 402 or 602) => Validate holidays,
- check you have some agents in site 1 and some in site 2,
- in holiday or comptime modules:
  - in the select of agents of the list page, you should see only agents in site 1,
  - you should see only existing holidays/comptimes for agent in site 1.
  - in the new form, you should be able to select agents in site 1 only.

### DIFF
--- a/src/Controller/CompTimeController.php
+++ b/src/Controller/CompTimeController.php
@@ -70,27 +70,8 @@ class CompTimeController extends BaseController
         // Search agents
         $agents = array();
         if ($this->admin) {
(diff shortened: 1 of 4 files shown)


jeromecombes (approved):

alexarn:
Merged on master and 21.04.xx

PR #421, opened 2021-04-26
MT30755: Add more working hours cycles

veggiematts:
Features:

 - nb_semaine configuration option now allows a cycle up to 10 weeks (3 before)
 - when using multi-sites, a number of weeks can be defined for each site
 - the working hour page displays the number of tables according to the sites
   the agent is affected to, or the default number of tables if nothing is
   defined for those sites
 - the user can choose from the different cycles available for the given agent
 - when adding working hours in multi-sites and there is only one site for this
   agent, pre-select this site.


alexarn:
Merged on master

PR #423, opened 2021-04-28
MT31714 Fix misuse of join function + replace join function with implode function

jeromecombes:


alexarn (approved):

alexarn:
Merged on master and 21.04.xx

PR #429, opened 2021-05-05
MT32677 : Don't show SR on grey cells, don't show right arrow on grey cells

jeromecombes:
MT32677 : Don't show SR on grey cells, don't show right arrow on grey cells


alexarn:
Merged on master and 21.04.xx

PR #435, opened 2021-05-06
MT32692 Prevent skills to be deleted if used by agents (20.10.xx only)

jeromecombes:
Prevent skills to be deleted if used by agents


alexarn:
Merged on 20.10.xx

PR #438, opened 2021-05-15
MT32775 : ICS parser gets the right date when the TimeZone contains ":"

jeromecombes:
MT32775 : ICS parser gets the right date when the TimeZone contains ":"
Also move cron.ics.php from public/ics to src/Cron/Legacy


alexarn:
Merged on master and 21.04.xx

PR #444, opened 2021-05-26
Redirect on route /agent after agent deletion

alexarn:
Related to MT 32675.
This bug seems only happen on firefox.

### DIFF
--- a/public/personnel/suppression.php
+++ b/public/personnel/suppression.php
@@ -81,6 +81,7 @@ function etape2()
 function etape3()
 {
     global $id;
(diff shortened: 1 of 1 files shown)


jeromecombes:
This PR does not work for 20.10.xx because the route /agent with method GET does not exist
Please use index.php?page=personnel/index.php for 20.10.xx

jeromecombes (approved):
Approved for master and 21.04.xx
Please use #447 for 20.10.xx

alexarn:

PR #446, opened 2021-05-27
Make installer ignore default $HOME/.my.cnf

alexarn:
This makes the mysql commands in install.sh file ignore local .my.cnf file.

Let's admit you have a $HOME/.my.cnf like the following:

```
[client]
user=plb
password=plb
host=localhost

[mysql]
default-character-set=utf8
database=planningBiblio
```

 - use the install.sh file with planningBiblio as wanted DB name.
 - the installer delete planningBiblio database => Ok
 - it try to re-create it => KO. This because the mysql command try to connect to the default database specified in .my.cnf. But it no longer exists.


jeromecombes (approved):

alexarn:

PR #451, opened 2021-06-02
Symfonize login and logout process

alexarn:


veggiematts (commented):
Tested ok with SQL login/logout.

However, with CAS enabled, I get:
Error: Internal script failure

veggiematts:
> However, with CAS enabled, I get:
> Error: Internal script failure

It works ok with an explicit port in CAS-Port and no protocol in CAS-Hostname. The documentation could be improved, but that is not the purpose of this PR. Review ok.

veggiematts (approved):

PR #454, opened 2021-06-03
MT 32845: Prevent user to change its own password via /myaccount with demo mode enabled

alexarn:


jeromecombes (approved):

alexarn:
Merged on 20.10.xx

PR #457, opened 2021-06-14
MT 32844: Compensatory time regularization

alexarn:


jeromecombes (approved):

veggiematts:
Warning: public/setup/atomicupdate/20210521_MT32844.php is adding Conges-fullday-reference-time but not Conges-fullday-switching-time

An updated installation will be missing this configuration option.

PR #467, opened 2021-07-01
Fix day models import

alexarn:
Test plan:
 - Create a planning on monday,
 - save it as day model (not week),
 - import this model on another day of the week,
 => planning stays empty.
 - apply this pull request and repeat the test plan.


jeromecombes (approved):

alexarn:
Merged on master and 21.04.xx

PR #469, opened 2021-07-13
MT 33011 - Replace HTML select with jquery timepicker input

alexarn:


public/setup/atomicupdate/MT33011_timepicker.php
```diff
@@ -1,2 +1,3 @@
 <?php
-$sql[]="UPDATE `{$dbprefix}config` SET valeurs = '[[1, \"Libre\"],[60,\"Heure\"],[30,\"Demi-heure\"],[15,\"Quart d&apos;heure\"],[5,\"5 minutes\"]]' WHERE nom = 'Granularite'";
+$sql[]="UPDATE `{$dbprefix}config` SET valeurs = '[[1, \"Libre\"],[60,\"Heure\"],[30,\"Demi-heure\"],[15,\"Quart d\'heure\"],[5,\"5 minutes\"]]' WHERE nom = 'Granularite'";
+$sql[]="UPDATE `{$dbprefix}config` SET valeur = '1' WHERE nom = 'Granularite'";
```
jeromecombes, line 3: Please remove this line.
We don't want to change this config for current users, but only for new installations.

jeromecombes (changes requested):

public/setup/atomicupdate/MT33011_timepicker.php
jeromecombes: Done on commit 6a741d4ab5715ef8b7f2a6ae86b4a621e2f9b6b9

jeromecombes (approved):

PR #472, opened 2021-07-21
Symfonyse planning import save delete

alexarn:


dogirod:
Issue when suppressing a chart.
Steps to reproduce :
1/ Import a model chart
2/ Lock the chart
3/ Refresh the page
4/ Unlock the chart
5/ Delete the chart
--> the page doesn't stop charging
I cannot figure out if it is due to my settings

alexarn:
All is in PR #466. So close

PR #474, opened 2021-08-11
MT29617: Multiple agents for holidays

veggiematts:
This patch allows to select several agents when adding a holiday (like when
adding an absence).

If the user doesn't have any of the following rights:
    "Gestion des congés, validation niveau 1"
    "Gestion des congés, validation niveau 2"
=> The selected user is the logged-in user, and this is not editable.

If the user has at least one of the following rights:
    "Gestion des congés, validation niveau 1"
    "Gestion des congés, validation niveau 2"
=> The user can choose one (and only one) agent

If the user has at least one of the following rights:
    "Gestion des congés, validation niveau 1"
    "Gestion des congés, validation niveau 2"
and the configuration option "Conges-Recuperations" is set to "Dissocier":
=> The user can choose one or more agents (but not all agents at the same time)

If the user has at least one of the following rights:
    "Gestion des congés, validation niveau 1"
    "Gestion des congés, validation niveau 2"
and the configuration option "Conges-Recuperations" is set to "Dissocier"
and the configuration option "Conges-tous" is enabled:
=> The user can choose one or more agents, or all agents at the same time.

This patch also adds a per site filter. All sites are checked by default.


alexarn:
Found one issue:
  => Select only one agent in the holiday form,
  => Choice dates from 18/08/2021 to 20/08/2021
  => Displayed hours is 21 and days 3 => Ok
  => Add another agent. displayed hours and days don't change => Not OK
  => Change the date(s). displayed hours and days don't change => Not OK

jeromecombes:
Found new issues,
Please see https://suivi.biblibre.com/view.php?id=29617#c190168

PR #496, opened 2021-09-23
Use a notifier as service container

alexarn:


veggiematts:
For a more concise syntax, couldn't we have a default transporter (CJMail in our case), leading to calls like:
```
$notifier->setRecipients($mail) [...]
```
instead of :
```
$notifier->setTransporter(new \CJMail())->setRecipients($mail) [...]
```

src/Controller/AgentController.php
```diff
@@ -869,21 +869,21 @@ public function save(Request $request)
...
-                $message.= "<ul><li>Login : $login</li><li>Mot de passe : $mdp</li></ul>";
-
-                $m = new \CJMail();
-                $m->subject = "Création de compte";
-                $m->message = $message;
-                $m->to = $mail;
-                $m->send();
+                $notifier = $this->container->get('notifier');
+                $notifier->setTransporter(new \CJMail())
+                         ->setRecipients($mail);
```
veggiematts, line 874: Is the semi-colon a typo ?

alexarn:
Merged on master

PR #506, opened 2021-10-21
MT 32769

jeromecombes:


src/PlanningBiblio/Helper/HolidayHelper.php
```diff
@@ -121,10 +122,41 @@ public function getCountedHours()
 
             $debutConges = $current == $debut ? $hre_debut : "00:00:00";
             $finConges = $current == $fin ? $hre_fin : "23:59:59";
+
+            // Convert free break as fixed break if needed
+            $free_break_already_removed = false;
+            $has_fixed_break = isset($planning['times'][$day_id][1]) && $planning['times'][$day_id][1] != "";
```
jeromecombes, line 128: You should also check $planning['times'][$day_id][5] : second pause when the setting pause2 is enabled.

Can this test be replaced by $has_fixed_break = !empty($planning['times'][$day_id][1]); ?

src/PlanningBiblio/Helper/HolidayHelper.php
```diff
@@ -121,10 +122,41 @@ public function getCountedHours()
...
+
+            if (isset($planning['breaktimes']) && !$has_fixed_break) {
+
+                $free_break_already_removed = true;
+                $free_break_start = $this->config('PlanningHebdo-DebutPauseLibre');
+                $free_break_end = $this->config('PlanningHebdo-FinPauseLibre')
```
jeromecombes, line 134: ; is missing

PR #510, opened 2021-10-30
Do not display hidden config elements

jeromecombes:
Allow to store hidden elements in config table (e.g. : contract information)
For master, 21.04 and 20.10


alexarn:

PR #511, opened 2021-11-03
Prevent creating empty working hours for non authorized users

alexarn:
Test plan:
  - Disable PlanningHebdo-Agents,
  - Log in with an agent without right for working hours,
  - Access /workinghour/add
  => Access is OK (http code 200) and should not,
  => You cannot edit hours (disabled) but you can save empty hours and should not.


alexarn:
Merged on master

PR #518, opened 2021-11-23
Initialize the widget calendar with the asked date when using the week view

jeromecombes:
Initialize the widget calendar with the asked date when using the week view
21.04.xx


alexarn:
Merged on 21.04.xx

PR #523, opened 2021-11-23
Fix update of dropdown menus in position form

jeromecombes:
TEST PLAN
On the position form, try adding items in the floor and group drop-down menus


jeromecombes:
Fixed by commit a15d235c58fc6d17537ff309cf0fe9ad4d76a53b on master and 21.10.xx

PR #527, opened 2021-12-01
Prevent reset of start/end hours if conges-heures is enabled

alexarn:


alexarn:
Test plan:
- Conges-mode = 'jour',
- Conges-heures enabled,
- You may need too add manually the conges-heures option in database (search the query in db_data.php),
- create a holiday with a start hour == 14h, end hour == 18h,
- save,
- edit this holiday
=> Start and end hours have disappeared

PR #530, opened 2021-12-07
Use url() js function for baseURL usage

alexarn:
See MT 35059


alexarn:
Merged on master and 21.10.xx

PR #537, opened 2021-12-28
MT35330 Fix redirection loop where CAS-ServiceURL is defined

jeromecombes:
MT35330 Fix redirection loop where CAS-ServiceURL is defined
Version 2, replace #536 

- sanitize $_GET params
- Isolate $_GET['ticket']


alexarn:
Merged on master and 21.10.xx

PR #543, opened 2022-01-08
MT35426_6: Allow edition of imported working hours

jeromecombes:
- Please see https://suivi.biblibre.com/view.php?id=35426#c199783


alexarn:
Merged on master and 21.10.xx.
This fix is already on branch BSG_21.10.xx_20220108 which is intended to be merge on BSG_21.10.xx

PR #544, opened 2022-01-08
MT35426_9 Fix display of "All site" on working hours on read only mode

jeromecombes:
- Also MT35426_10 Display hours in plain text instead of input disabled
 - Please see https://suivi.biblibre.com/view.php?id=35426#c199790


jeromecombes:
Replaced with #560

PR #545, opened 2022-01-08
MT35426_2 : defined variables on WorkingHoursController

jeromecombes:
- Please see https://suivi.biblibre.com/view.php?id=35426#c199775


jeromecombes:
Merged on MT35426_9 #544 to avoid conflicts

PR #548, opened 2022-01-12
Take halfday into account in holiday controller

alexarn:
Test plan in MT 35277


alexarn:
Merged on master and 21.10.xx

PR #550, opened 2022-01-17
Remove PHP Notice when fetching positions

alexarn:
See MT 35505


jeromecombes (approved):
testé et approuvé

alexarn:
Merged on master and 21.10.xx

PR #553, opened 2022-01-21
Show sites on the read-only agents page

alexarn:
Test plan:
  - Multi-sites enabled,
  - Give an agent the "See agent account" only
  - Login with this agent,
  - Show an agent page
  => Sites are not shown
  - apply this patch


jeromecombes (approved):
revue et tests : OK

alexarn:
Merged on master 21.10.xx and 21.04.xx

PR #557, opened 2022-01-24
MT35523 Fix the "Call For Help" mail function #549 (for master)

jeromecombes:
MT35523 Fix the "Call For Help" mail function #549
Please see https://suivi.biblibre.com/view.php?id=35523


alexarn:
Merged on master

PR #558, opened 2022-01-27
Remove spaces from some read-only values in working hours table.

alexarn:
These spaces cause bad duration calculation.
**21.04.xx only**


dogirod:
Tested --> ok - the spaces are withdrawn and the calculation is ok.

alexarn:
Merged on 21.04.xx

PR #559, opened 2022-01-27
Adjust JS function plHebdoMemePlanning to new timepicker widget

alexarn:
Test plan:
  - PlanningHebdo On, Multi-Site On,
  - Create working hours with different values on 2 weeks ut with the same site on all line,
  - On read-only mode (agent acount) there is only the first week displayed.


jeromecombes (approved):
revue : OK
tests : OK
Approuvée

alexarn:
Merged on master and 21.10.xx

PR #561, opened 2022-01-28
Use dependency injection instead of service container to get Notifier into controllers

alexarn:
Since we use Symfony 4.4, our controllers extends AbstractController
instead of deprecated Controller Bundle. Now we can't retrieve our custom
services directly from container.


alexarn:
Merged on master

PR #562, opened 2022-01-28
MT35413: Restrict absences list interval to a year maximum

veggiematts:
Test plan:
 - set an interval superior to a year in the absences list parameters
 - check that a message is displayed to tell you to select an interval inferior than a year
 - check that it also works when only the start date or end date is set

### DIFF
--- a/public/absences/js/voir.js
+++ b/public/absences/js/voir.js
@@ -33,6 +33,24 @@ $(function(){
     });
   });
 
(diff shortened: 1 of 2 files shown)


jeromecombes (approved):
revue et tests : OK
Approuvée

alexarn:
Merged on master and 21.10.xx

PR #563, opened 2022-01-28
Avoid comparison between H:i:s and H:i when checking the presence of an agent

alexarn:
Test plan:
  - Set working hours for agent like: 10h00, 13h00, 14h30, 17h30
  - Try to place it on a planning for a position starting at 10h ending at 11h
  => Ok
  - Try to place it on a planning for a position  ending at 13h
  => Ko because of comparison between 13:00:00 and 13:00


alexarn:
Merged on master and 21.10.xx

PR #565, opened 2022-02-01
Don't reload holiday form for multiple selection

alexarn:


alexarn:
Merged on master and 21.10.xx

PR #575, opened 2022-02-09
Check if free breaktime is not empty in agent form

alexarn:
See MT 35757


jeromecombes (approved):

alexarn:
Merged on master and 21.10.xx

PR #583, opened 2022-03-01
MT35991: Fix AbsenceInfoController to use DateTime

veggiematts:
MT34252 switched start and end date of absences infos from string to date in
model. Absence infos were broken since.

Test plan:
 - Try adding or updating an absence info, and check that it doesn't work (500 error)
 - Apply the patch
 - Check that listing, adding and editing absence info now works.


jeromecombes (approved):
revue et tests : OK
à merger sur master (mais pas sur 21.04, ni sur 21.10)

alexarn:
Merged on master

PR #584, opened 2022-03-01
MT35774: Fix error on free break calculation

veggiematts:
Holiday calculation can cause a crash when APP_DEBUG=1


alexarn:
Merged on master and 21.10.xx

PR #588, opened 2022-03-08
Don't show all agents plannings on absence add notification

alexarn:


alexarn:

PR #589, opened 2022-03-08
MT36087: fix encoding issues on agents' name in calendar's dropdown menu

jeromecombes:
- When agent's name contains apostrophes
- This fix is relative to M36087 but does not fix the client request


alexarn (approved):

alexarn:
Merged on master, 21.10.xx and 21.04.xx

PR #592, opened 2022-03-10
Show all site in ready-only agents working hours tab

alexarn:
Related to MT 36131


alexarn:
Merged on master and 21.10.xx

PR #594, opened 2022-03-11
MT34284: Handle exception on recurring events with the recurrence-id attribute

jeromecombes:
MT34284: Handle exception on recurring events with the recurrence-id attribute


alexarn:
Merged on master, 21.10.xx and 21.04.xx

PR #596, opened 2022-03-15
Prevent adding working hour exception out of parent's date range

alexarn:
Related to MT 36157


jeromecombes (approved):
revue et tests : OK
approuvée

alexarn:

PR #597, opened 2022-03-21
MT35648: Optimize DB queries for holiday/credits (21.04.xx)

jeromecombes:


public/conges/class.conges.php
```diff
@@ -551,7 +551,7 @@ public function fetchAllCredits()
         // Crédits actuels
         // Sélection des derniers congés validés
         $db=new db();
-        $db->select("conges", null, "valide>0", "ORDER BY `validation` desc");
+        $db->select("conges", null, array("valide>0", "validation>$debut"), "ORDER BY `validation` desc");
```
alexarn, line 554: The third paramter of db select method must be a string not an array. I get this PHP notice when loading the holiday credits:
`Array to string conversion in /home/plb/planningBiblio/public/include/db.php on line 130`

And so, it is missing lot of values in the table.

Either the third paramter be a string or you can use the same method of dbh class (That handle string and array).

jeromecombes: Thanks @alexarn ,

I replaced db::select with db::select2 that take an array as third parameter.
There's no error anymore, but I realized that the result is not correct.
I have to find out another solution.

DO NOT MERGE

jeromecombes:
Replaced with #616

jeromecombes:
Replaced with https://github.com/PlanningBiblio/PlanningBiblio/pull/616

PR #599, opened 2022-03-21
MT35648: Optimize DB queries for holiday/credits (master)

jeromecombes:


public/conges/class.conges.php
```diff
@@ -577,7 +577,7 @@ public function fetchAllCredits()
         // Crédits actuels
         // Sélection des derniers congés validés
         $db=new db();
-        $db->select("conges", null, "valide>0", "ORDER BY `validation` desc");
+        $db->select("conges", null, array("valide>0", "validation>$debut"), "ORDER BY `validation` desc");
```
alexarn, line 580: The third paramter of db select method must be a string not an array. I get this PHP notice when loading the holiday credits:
`Array to string conversion in /home/plb/planningBiblio/public/include/db.php on line 130`

And so, it is missing lot of values in the table.

Either the third paramter be a string or you can use the same method of dbh class (That handle string and array).

public/conges/class.conges.php
```diff
@@ -598,7 +598,7 @@ public function fetchAllCredits()
         // Crédits actuels
         // Sélection des dernières mises à jour de crédits
         $db=new db();
-        $db->select("conges", null, "information>0", "ORDER BY `info_date` desc");
+        $db->select("conges", null, array("information>0", "info_date>$debut"), "ORDER BY `info_date` desc");
```
alexarn, line 601: Same here

public/conges/class.conges.php
```diff
@@ -623,7 +623,7 @@ public function fetchAllCredits()
         // Crédits actuels
         // Sélection des dernières mises à jour de récup
         $db=new db();
-        $db->select("recuperations", null, "valide>0", "ORDER BY `validation` desc");
+        $db->select("recuperations", null, array("valide>0", "validation>$debut"), "ORDER BY `validation` desc");
```
alexarn, line 626: Same

alexarn (changes requested):

jeromecombes:
Holliday credits are not calculated correctly.

PR #600, opened 2022-03-21
MT 36224: Get more active menus

alexarn:
Add a twig filter that check for specific active menus:
  - Admin pages,
  - Holiday pages

Also, update that check for non symfonized pages:
  - planning/volants/index.php
  - congés/*


jeromecombes (approved):
revue et tests : OK

alexarn:
Merged on master

PR #601, opened 2022-03-23
Conditionally transform times on 2 digits in function calculSiPresent

alexarn:
See MT 36259


jeromecombes (approved):

alexarn:
Merged on master and 21.10.xx

PR #603, opened 2022-03-31
MT36346: Allow symfony-flex plugin without confirmation

jeromecombes:
MT36346: Allow symfony-flex plugin without confirmation

Without this commit, when we run "composer install" or "install.sh", we need to confirm the installation of symfony flex.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,10 @@
         "preferred-install": {
             "*": "dist"
         },
-        "sort-packages": true
+        "sort-packages": true,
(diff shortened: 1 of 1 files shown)


alexarn:
Merged on 21.10.xx

alexarn:
Merged on 21.10.xx with one more commit:
https://github.com/PlanningBiblio/PlanningBiblio/commit/7432271fcabdebcf4fa1cfe8093de01ee7d1634f

PR #605, opened 2022-03-31
MT36346: Allow symfony-flex plugin without confirmation

jeromecombes:
MT36346: Allow symfony-flex plugin without confirmation

Without this commit, when we run "composer install" or "install.sh", we need to confirm the installation of symfony flex.


alexarn:
composer.json rebuilt on 21.04.xx

PR #608, opened 2022-04-06
MT36370: Do not filter on first site when gathering absents for statistics

veggiematts:
On the attendees/missing statistics page, some agents may not appear as absents although they are.
    
This is because the list of absents is built with the first site hardcoded on fetch call.
    
This patch removes the filter to have all agents.


alexarn (approved):

alexarn:

PR #617, opened 2022-04-08
MT36452 - Prevent Hours helper to convert non H:i values

alexarn:
Recently merged PR #587 brought a new method to convert H:i values to H:i:s. Problem, sometimes non H:i values are passed to it causing a fatal error. Test plan:

- Enable multi-sites,
- enable PlanningHebdo,
- Edit workinghours and set some sites in the table.
- Save
   => Crash: PHP Fatal error:  Uncaught Error: Call to a member function format() on bool 

This also cause a fail when updating to 21.11.00.002.

### DIFF
--- a/src/PlanningBiblio/Helper/HourHelper.php
+++ b/src/PlanningBiblio/Helper/HourHelper.php
@@ -40,6 +40,10 @@ class HourHelper extends BaseHelper
             return '';
         }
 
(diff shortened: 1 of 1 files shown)


veggiematts (approved):

alexarn:
Merged on master and 21.10.xx

PR #622, opened 2022-04-19
Follow up: Allow to load absences without documents for performance i…

jeromecombes:
…mprovement

See https://www.planningbiblio.fr/forums/topic/lenteurs-lors-de-la-recuperation-des-agents-disponible/#post-3264
and https://suivi.biblibre.com/view.php?id=36408#c206678


alexarn (approved):

alexarn:

PR #624, opened 2022-04-20
MT36552: Show information that end today in absence information list

jeromecombes:
MT36552: Show information that end today in absence information list

See MT36552


alexarn (approved):

alexarn:
Merged on master and 22.04.xx (last commit only).
Not merged on 21.10.xx

PR #625, opened 2022-04-20
MT36550: Show agent's name in absence form even if he is deleted

jeromecombes:
MT36550: Show agent's name in absence form even if he is deleted

See MT36550


dogirod:
I see the agent in the pick list but not on the absence details page.

alexarn:

PR #626, opened 2022-04-20
MT36561: Check if [0][0] exists in the function calculSiPresent

jeromecombes:
See MT 36561


jeromecombes (commented):
revue et tests : OK
Approuvée

alexarn:
Merged on master, 22.04.xx and 21.10.vv

PR #627, opened 2022-04-20
MT36556: Display all grey cells in the same way

jeromecombes:
See MT36556


alexarn:
Merged on master and 22.04.xx

alexarn:
Merged on master and 22.04.xx

PR #628, opened 2022-04-20
MT36558: Correctly hide and show drop, import and print icones in the…

jeromecombes:
… Planning menu

see MT36558


alexarn:
Merged on master and 22.04.xx
Rebased for 21.10.xx

PR #635, opened 2022-04-22
AFI456: Planook: Add Planook configuration

jeromecombes:
AFI456: Planook: Add param Planook, hide some information in Agent's form

AFI456: Planook: hide some menus with condition "config!=Planook"

AFI456: Planook: hide the button for changing Absence Reasons

AFI456: Replace the default empty string with an empty json array for agent skills

- Changing the default value for agent skills allows agents to be selected on positions that do not require any skill.

AFI456: Hide some information on position templates

AFI456: Hide the option "Change selection" on agent list when Planook is enabled

AFI456: Add the SQL queries to run to enable Planook config


alexarn (changes requested):
Page /agent is on error with APP_DEBUG=1:
Undefined variable: postes_completNoms on src/Controller/AgentController.php (line 186)
Probably because there is no longer skills
IMO this could happen on master if users delete all skills

Contextual menu does not appear on planning:
Undefined offset: 0" at /home/planningb/www/planningbiblio/public/include/function.php line 604
Happen when no agent has working hours.

jeromecombes:
> > Contextual menu does not appear on planning:
> > Undefined offset: 0" at /home/planningb/www/planningbiblio/public/include/function.php line 604
> > Happen when no agent has working hours.
> 

Fixed by rebase with 22.04.xx

PR #641, opened 2022-04-27
MT 36558: followup - allow all user to see print icon

alexarn:


alexarn:

PR #642, opened 2022-04-27
ICS import: Allow custom exclusions from custom_options.php

jeromecombes:
This commit allows the creation of a custom function that can be used to filter events before importing them
In the optional custom_options.php file, you can add the $config['ICS-custom-exclusion'] parameter and the associated function.

E.g. :
$config['ICS-custom-exclusion'] = 'myExclusionFunction';

function myExclusionFunction($event) {
    return $event['SUMMARY'] != 'Absent';
}

// Events that do not match with "Absent" won't be imported

### DIFF
--- a/public/ics/class.ics.php
+++ b/public/ics/class.ics.php
@@ -200,6 +200,14 @@ class CJICS
 
         $events=array();
         foreach ($tmp as $elem) {
+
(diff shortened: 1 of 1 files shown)


alexarn:
Merged on master

PR #643, opened 2022-04-28
Holiday information : remove HTML entities and display information dates

jeromecombes:
+ optimize holiday form and drop down menu widths

TEST PLAN : 
- Add an holiday information with simple and double quotes
- Add a new holiday
- --> Information dates are now displayed
- --> no html entities is displayed


alexarn:
Merged on 21.10.xx

PR #645, opened 2022-04-29
Prevent editing used frameworks only if someone is linked to them

jeromecombes:


alexarn:

PR #648, opened 2022-05-03
MT36715: fix import working hours

jeromecombes:
Since Release 21.10.00, working hours can be stored with the nb_semaine parameter, which can be different from config['nb_semaine'].
The import script did not set this parameter, so the imported working hours was always displayed over a week, even if config['nb_semaine'] was higher.
The import script managed rotation of working hours on different weeks by adding an offset to the day (+7, +14, etc.), so some hours were not visible in the web interface because only the first week was displayed

The modified script sets nb_semaine = 1 and does not add the offset to the day. So the hours are always visible.

### DIFF
--- a/src/Cron/Legacy/cron.importCSV.php
+++ b/src/Cron/Legacy/cron.importCSV.php
@@ -1,40 +1,27 @@
 <?php
 /**
-Planning Biblio, Version 2.7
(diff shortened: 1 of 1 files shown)


alexarn:

PR #650, opened 2022-05-06
MT36683 : Fix lack of css when a new reason for an abscence was added

alexarn:


alexarn:
Merged on master and 22.04.xx

PR #651, opened 2022-05-06
MT36684 : Fix radio buttons alignement in the choice of timetables fo…

alexarn:
…r the week in agent/workinghours


jeromecombes (approved):
revue et tests : OK
approuvée

alexarn:
Merged on master and 22.04.xx

PR #653, opened 2022-05-09
MT 36809: prevent include class.conges.php twice

alexarn:
This fix a fatal error when changing comp-time status.


jeromecombes (approved):

alexarn:

PR #654, opened 2022-05-09
MT 36248: Disabled notifications should not be considered as an error

alexarn:
If the notification system is disabled, CJMail class consider this as an error. This cause some empty error message in some places.

I.e: create an absence for more than one agent.


jeromecombes (approved):
revue OK
tests OK
PR approuvée

alexarn:

PR #659, opened 2022-05-13
Use the granularity setting for planno-break-timepicker

jeromecombes:


alexarn:

PR #660, opened 2022-05-16
Transform working hours to His format in agents table

alexarn:


src/PlanningBiblio/Helper/HourHelper.php
```diff
@@ -51,4 +51,8 @@ public static function toHis($time)
 
         return $time;
     }
+
+    public function WorkingHoursToHis($workinghours)
```
veggiematts, line 55: Why ?

veggiematts (changes requested):
Seems ok apart from the empty WorkingHoursToHis function

veggiematts (approved):

PR #661, opened 2022-05-24
MT36802 : Do not set SMTP settings when the local mailer is used

jeromecombes:
Please see MT36802
test sending mails with param Mail-IsMail-IsSMTP = IsMail


alexarn:

PR #670, opened 2022-06-10
MT36303:  Add more precision to hours count when adding holiday

veggiematts:
Test plan:
    
     - Find a configuration that leads to an imprecision in the number of hours
       calculation when adding an holiday. It is likely to happen when an holiday
       is long enough. Dynamic number of hours calculation on the holiday page
       will lead to something like 56h41 (instead of 56h40).
    
     - Apply this patch.
    
     - Check that the amount of hours is now correct.


jeromecombes (approved):
revue et tests : OK
PR approuvée

alexarn:

PR #671, opened 2022-06-10
Mt37151

jeromecombes:
- Do not convert comp-time in days on My Account (Comp-time is always in hours)
- Do not display credits in hours on My Account if Conges-Mode=jours
- Correctly calculate holiday credits in days on My Account
- Use $fulldayReferenceTime variable to calcule credits in days, Its default value is 7 hours, it takes config(fullday_reference_time) if defined, and it is always 7 hours if Conges-Mode=jours

NB: if Conges-Mode=jours, credits are stored and computed in hours, then converted in days for display, always with 7 hours as reference time. Config(fullday_reference_time) is used to display the equivalent in days when credits are managed in hours.

### DIFF
--- a/src/Controller/AccountController.php
+++ b/src/Controller/AccountController.php
@@ -54,16 +54,25 @@ class AccountController extends BaseController
 
         // Crédits (congés, récupérations)
         if ($this->config('Conges-Enable')) {
+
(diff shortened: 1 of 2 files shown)


alexarn:
Merged on master and 22.04.xx

PR #690, opened 2022-07-13
Remove absence class from time-slot and make them dropable

alexarn:
Test plan:
In the stated week page, place some agents like this:
  - In the first (left) column:
    - an agent whose name starts with A....
    - an agent whose name starts with C.... and make he is absent (red crossed)
  - in the second (middle) column:
    - an agent whose name starts with B....
  - move the B.... agent from middle to left column
  => The B... agent become absent.

Apply this patch and try again.


jeromecombes (approved):
revue de code : OK

alexarn:
Merged on BSG_master_20220609

PR #693, opened 2022-07-28
Add Planook config for 22.04.xx

jeromecombes:
Replace #635 

PR for 22.04.xx

@alexarn 
From #635 

> Page /agent is on error with APP_DEBUG=1:
> Undefined variable: postes_completNoms on src/Controller/AgentController.php (line 186)
> Probably because there is no longer skills
> IMO this could happen on master if users delete all skills
--> Fixed by commit 449466a9ed87e59ed6644ee2baa40dc07f85a310

> Contextual menu does not appear on planning:
> Undefined offset: 0" at /home/planningb/www/planningbiblio/public/include/function.php line 604
> Happen when no agent has working hours.
> 

--> Fixed by rebase with 22.04.xx


alexarn:
Merged on 22.04.xx

PR #694, opened 2022-07-28
Add Planook config for master

jeromecombes:
Replace #635 

PR for master

@alexarn 
From #635 

> Page /agent is on error with APP_DEBUG=1:
> Undefined variable: postes_completNoms on src/Controller/AgentController.php (line 186)
> Probably because there is no longer skills
> IMO this could happen on master if users delete all skills
--> Fixed by commit b8240a4f780d6d90c99133cc2bf42f754f7e6cb9

> Contextual menu does not appear on planning:
> Undefined offset: 0" at /home/planningb/www/planningbiblio/public/include/function.php line 604
> Happen when no agent has working hours.
> 

--> Fixed by rebase with master


alexarn:
Merged on master

PR #697, opened 2022-08-03
PHP 8.1 compatibility

jeromecombes:


jeromecombes:
@alexarn Salut, peux-tu ajouter un job Circle-CI pour PHP 8.1 ?

alexarn:
The 3 builds are green. Ok. But removing the line that add the coverage mode for xdebug make we lost the code coverage reports.
I don't realy understand why, in the new used PHP images, the $PHP_INI_DIR variable point to a directory whithout php.ini file.

jeromecombes:
> The 3 builds are green. Ok. But removing the line that add the coverage mode for xdebug make we lost the code coverage reports. I don't realy understand why, in the new used PHP images, the $PHP_INI_DIR variable point to a directory whithout php.ini file.

@alexarn Codecov is restored with the commit 8fe2f35647184f62a2b1ff74739ed6215d9bf238 
I installed the xdebug driver and added "xdebug.mode=coverage" to the file /etc/php.d/circleci.ini (the only php.ini file which is loaded on the new php images)

alexarn:
Merged on master and 22.04.xx

PR #708, opened 2022-08-26
MT 37712: define used variable to false when adding new framework

alexarn:
Test plan in MT 37712


jeromecombes (approved):
revue et tests : OK
PR approuvée

alexarn:
Merged on master and 22.04.xx

PR #712, opened 2022-08-30
MT 37839: fix the query that remove duplicates in DBRev 22.04.00.000

alexarn:
I fixed the query in the related DBRev and also added a new one for people having already executed the 22.04.00.000


jeromecombes (approved):
Revue et tests : OK
approuvée

alexarn:
Merged on master and 22.04.xx

PR #723, opened 2022-09-13
MT37943: MSGraph sync: Fix update event query and improve logs

veggiematts:
Previously, the update event query could update the wrong absence when multiple absences had the same ical_key (due to the lack of perso_id in the query)


jeromecombes (approved):
revue : OK

PR #726, opened 2022-09-15
MT 37442: check if agents are working when deleting absences

alexarn:


jeromecombes (changes requested):
Voir mes notes sur le ticket

PR #740, opened 2022-09-30
MT38395 : fix check mail config

jeromecombes:
see MT38395

TEST PLAN : 
- edit config
- change mail settings and click the button "Tester" each time you change a field
- the test should take values from the fields before the form is submitted, and show a success or error result depending these values


alexarn (approved):

alexarn:
Merged on master and 22.04.xx

PR #764, opened 2022-10-31
MT38746: Fix wrong URLs

jeromecombes:
Please see https://suivi.biblibre.com/view.php?id=38746 and https://www.planno.fr/forums/topic/bug-modif-poste-renvoi-sur-mauvaise-url/


veggiematts (approved):

alexarn:
Merged on master and 22.10.xx

PR #770, opened 2022-11-10
MT38868: Fix quota comparison

jeromecombes:
Please see MT38868


alexarn:
Merged on master and 22.04.xx

PR #771, opened 2022-11-15
Remove phpunit direct depreciations

jeromecombes:


alexarn:
Merged on master and 22.10.xx

PR #772, opened 2022-11-15
MT38918 : delete references to planning_hebdo_tables

jeromecombes:
see MT38918


jeromecombes (commented):
commit MT 38918: remove useless method getPeriodes et ensemble de la PR approuvés

alexarn (approved):

PR #807, opened 2022-12-08
Fix rowCounts error

qlouis1:
PurgeLogTable throws an error :

> In PurgeLogTable.php line 47:
>                                                                                              
>   Attempted to call an undefined method named "rowCount" of class "Doctrine\DBAL\Statement".

The script purges the logs from the database correctly but the error prevents it to log the execution.

Seems like rowCounts() must be called on the result of the query and not the Statement.


veggiematts:
Thanks for the feedback. I'm approving the PR.

PR #808, opened 2022-12-08
MT39248: prohibit changing the password in demo mode

jeromecombes:
MT39248: prohibit changing the password in demo mode

See MT39248


alexarn:
Merged on 22.10.xx

PR #816, opened 2022-12-13
MT39277: Include deleted agents in statistics by position

veggiematts:
Include deleted agents in statistics by position to prevent wrong calculation and attribution to 'ZZZ_Autres' status or service.


jeromecombes (approved):
revue et tests : OK
approuvée pour 22.04.xx

alexarn:
Merged on 22.04.xx

PR #821, opened 2022-12-20
MT37569: adapt holiday reseting cron jobs

jeromecombes:
- move db crons in src/Cron/Legacy
- add a dedicated reset cron for compensatory time
- conditionaly reset comp time in cron.holiday_reset_credits.php
- ability to disable cron
- code refactoring

This PR replace #806 (same PR but squashed)


alexarn:
Merged on master and 22.10.xx

PR #826, opened 2022-12-27
MT39408: Replace fetchAll with fetchAllAssociative in MSGraphClient

veggiematts:
$statement->fetchAll is deprecated and doesn't work on current DBAL/Doctrine version.

$results->fetchAllAssociative is used instead.

Also, remove deprecated rowCount calls which were used only for debug purposes.


jeromecombes (commented):
revue OK

jeromecombes (approved):
Revue OK.
Fonctionnement OK sur l'instance du client.

alexarn:
Merged on master, 22.10.xx and 22.04.xx

PR #834, opened 2023-01-06
MT39508: fix a migration error (duplicate entry CAS-LoginAttribute)

jeromecombes:


alexarn:
Merged on master and 22.10.xx

PR #837, opened 2023-01-11
Revue to avoid direct deprecation notices

alexarn:


jeromecombes:
Changes are already made on master

PR #856, opened 2023-01-18
MT39412: Use db->sanitize_string = true in maj.php and tools scripts

jeromecombes:
Since 22.10.04, the default value for db->sanitize_string is false. This is helpfull to print data with Symfony templates but this generates errors when we migrate old data.


alexarn:
Merged on master and 22.10.xx

PR #870, opened 2023-02-13
MT39895: Add throttle to overtime edits

veggiematts:
In some cases (network hiccups for example), the POST request for editing an overtime can be sent twice.

This patch adds a minimum delay between two edits of a given overtime of 1 second by default.

This value can be overridden with the following setting in custom_options.php:

$config['post_requests_throttle'] = n;

Test plan:

1) Set post_requests_throttle to a large enough value in custom_options.php
   (20 seconds should be enough)

2) Edit an overtime, check that the modification is made.

3) Within the 20 seconds, edit this overtime again, check that the modification
   is denied, with a message telling you to wait a few seconds before trying
   again.

4) After the 20 seconds are expired, edit this overtime again, and check that
   the modification is made.

### DIFF
--- a/src/Controller/OvertimeController.php
+++ b/src/Controller/OvertimeController.php
@@ -4,7 +4,9 @@ namespace App\Controller;
 
 use App\Controller\BaseController;
 use App\PlanningBiblio\Helper\HolidayHelper;
+
 use App\Model\Agent;
+use App\Model\OverTime;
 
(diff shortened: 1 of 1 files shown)


jeromecombes (approved):
revue et tests : OK
PR approuvée pour master, 22.10.xx et BSG

PR #875, opened 2023-02-17
MT39548: Fix models deletion

veggiematts:
Test plan:

1) Without the patch, save a model

2) Save the model again, with the same name

3) In models administration, check that you have to delete it twice, which is
   not normal

4) With the patch, save a model

5) Save the model again, with the same name

6) In models administration, check that you have to delete it only once


jeromecombes (approved):
revue et tests : OK
approuvée pour master, 22.10.xx

PR #880, opened 2023-02-22
MT39696: Set Recup-DelaiDefaut to 0 if not correctly set.

veggiematts:
Test plan:

 1) Set Recup-DelaiDefaut (Conges section) to blank or to a non-integer value.

 2) Without the patch, check that a new overtime request triggers a JS error.

 3) With the patch, check that there is no JS error.
    (Recup-DelaiDefaut is set to 0)

### DIFF
--- a/src/Controller/OvertimeController.php
+++ b/src/Controller/OvertimeController.php
@@ -168,7 +168,7 @@ class OvertimeController extends BaseController
         }
 
         $this->templateParams(array(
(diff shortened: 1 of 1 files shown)


jeromecombes (approved):
revue et tests : OK
approuvée pour master et 22.10.xx
Ne s'applique pas sur 22.04.xx

PR #883, opened 2023-03-03
MT39529: Fix erroneous message on absence add

veggiematts:
Test plan:

 1) Without the patch, make sure that direct managers (Reponsables directs)
 are checked in at least one of the absences notifications workflows
 (Absences-notifications-* in the Absences tab of the configuration page)

 2) For a user that has no direct managers, add an absence

 3) Check that the message "`<li></li>`" is displayed before the message saying
 that the absence has been added.

 4) With the patch, check that the message "`<li></li>`" is gone.

This happens because there is an empty email in the list of recipients.

This patch fixes the problem in two ways:

 - Make get_manager_emails return an empty array when there are no managers (and not an array with an empty value)

 - Remove empty values in getRecipients* functions to make sure no empty emails are passed to the CJMail class.

### DIFF
--- a/public/absences/class.absences.php
+++ b/public/absences/class.absences.php
@@ -1244,7 +1244,7 @@ class absences
             }
         }
 
-        $this->recipients=$recipients;
+        $this->recipients = array_unique(array_filter($recipients));
     }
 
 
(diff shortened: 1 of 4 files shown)


jeromecombes (commented):
Revue OK
mais je reproduis toujours l'erreur

jeromecombes (approved):
revue et tests : OK
Approuvée pour master et 22.10.xx

Ne s'applique pas sur 22.04.xx

PR #884, opened 2023-03-06
MT39693: Show closingday in menu only if Conges-Enable is enabled

veggiematts:
Test plan:
    
     - Run the SQL update (public/setup/updatedatabase.php)
    
     - Check that "Jours de fermeture" (/closingday) is displayed in the
     Administration menu only if Planook is disabled and Conges-Enable is enabled.

### DIFF
--- a/public/include/class.menu.php
+++ b/public/include/class.menu.php
@@ -31,41 +31,48 @@ class menu
     {
     }
 
+    public function checkCondition($condition) {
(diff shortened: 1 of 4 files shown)


jeromecombes (approved):
revue et tests : OK
approuvée pour master, 22.10.xx, 22.04.xx

PR #891, opened 2023-03-22
New sites management

alexarn:


IPandragonI:
Hello, is there a reason why this solution was not implemented ?
Will this feature be merge one day ?
Thanks for your answer

PR #898, opened 2023-04-06
MT40167: Strike agents that have left in the stated week plannings.

veggiematts:
Test plan:

 1) Without the patch, place an agent in a stated week planning at a given date

 2) Edit that agent to set his leaving (depart) date at a prior date

 3) Check that the agent is not striked in the stated week planning

 4) Apply the patch, and check that the agent is striked in the stated week
    planning, with the following message: "Date de départ dépassée"

### DIFF
--- a/public/js/statedweek.js
+++ b/public/js/statedweek.js
@@ -296,6 +296,10 @@ function placeOnPlanning(agent) {
         cell.addHoliday();
       }
 
+      if (agent.leaving_date_passed) {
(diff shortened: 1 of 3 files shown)


jeromecombes (approved):
revue et tests : OK
Approuvée pour la branche BSG_master_20220725

jeromecombes:
Merged on BSG_master_20220725

PR #899, opened 2023-04-07
MT39836: interpret html tags on comments in weekly view

jeromecombes:
Please see MT39836


btangaro:
functional test : ok. 
The html line break code is no longer displayed

PR #900, opened 2023-04-11
MT39966: Fix comments display under plannings

veggiematts:
Test plan:

Without the patch:

 - Enable Planning-CommentairesToujoursActifs
 - Give only the "Modification des commentaires des plannings" to a user
 - With this user, try to add a comment to an unvalidated planning
 - Check that the "Ajouter un commentaire" button appears twice and has no effect.

With the patch:

 - In the same conditions, check that you can add a comment.

### DIFF
--- a/templates/planning/poste/index.html.twig
+++ b/templates/planning/poste/index.html.twig
@@ -47,7 +47,7 @@
       {% endif %}
     {% endif %}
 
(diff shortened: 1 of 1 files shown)


jeromecombes (approved):
revue et tests : OK
PR approuvée

PR #902, opened 2023-04-14
MT40668: ICS export with absences

jeromecombes:
MT40668: ICS export with absences
Please see MT40668


veggiematts:
Tested ok, code review ok.
This feature has no documentation ?
There should be at least a new link in the agent account for the export with absences.

jeromecombes:
replaced with #916 (squashed)

PR #907, opened 2023-04-25
MT40756: Restore popup function and agent deletion

jeromecombes:
MT40756: Restore popup function and agent deletion

Please see MT40756

/!\ 22.10.xx only


veggiematts (approved):

jeromecombes:
Merged on 22.10.xx

PR #918, opened 2023-06-07
MT41094: Don't check journey time between areas when there is only one area

jeromecombes:
MT41094: Don't check journey time between areas when there is only one area


LucyMichel:
test ok sur MT41094, on peut de nouveau accéder normalement à la liste des agents

veggiematts (approved):

PR #920, opened 2023-06-08
MT39946: Fix time format parameter for ajax call in absence add

veggiematts:
holidayAbsenceControl expects debut and fin parameters to be in the following format: hh:mm:ss

public/absences/js/modif.js passed them in a hh:mm format, causing wrong date comparison in SQL query.


jeromecombes (approved):
revue et tests : OK
approuvée pour master, 22.10.xx, 23.04.xx

PR #926, opened 2023-06-12
MT41169: Add the background color on Datatable's select

jeromecombes:
MT41169: Add the background color on Datatable's select


btangaro:
Functional test OK

veggiematts (approved):

PR #928, opened 2023-06-19
MT41068: Fix absence statistics display when PlanningHebdo is disabled

veggiematts:
Test plan:

 - Disable PlanningHebdo
 - Go to absence statistics (statistics/absence)
 - Check that there a more columns than columns headers
 - Apply the patch
 - Check that the number of columns match the number of columns headers


jeromecombes:
test avec le 2ème commit : OK

PR #929, opened 2023-06-19
MT40167: Strike agents that have left in the stated week plannings

veggiematts:
- Don't show an agent as available in context menu if his leaving date is
     before the given stated week planning.


jeromecombes (approved):
revue et tests : OK
approuvée pour BSG_master_20220725

jeromecombes:
merged on BSG_master_20220725

PR #930, opened 2023-06-26
MT41151: Allow /holiday/accounts to be displayed in days

veggiematts:
Currently, /holiday/accounts (Congés -> Crédits) is always displayed in hours regardless of the Conges-Mode parameter.

Test plan:

1) Without the patch, values are always displayed in hours, even if
   Conges-Mode is set to "jours"

2) With the patch, check that values are displayed accordingly to Conges-Mode
   setting.


jeromecombes (approved):
revue et tests : OK

PR #937, opened 2023-07-26
MT41650: Use utf8-aware function mb_substr (instead of substr) in plannings

veggiematts:
Test plan:

 - Without the patch, reloading a planning page having an already added agent with an utf8 first character in his first name (like an accent, "é" for example) will display the Unicode Replacement Character: �

 - With the patch, the agent first name will be correctly shortened ("Émile" will become "É." for example).

See Also:
MT41411: Use utf8-aware function mb_substr (instead of substr) in plannings #934


jeromecombes (approved):
revue et tests : OK
approuvée

jeromecombes:
force-push = rebase master

PR #940, opened 2023-08-24
MT41778: CircleCI: Download Google Chrome v114 from archive for compa…

jeromecombes:
MT41778: CircleCI: Download Google Chrome v114 from archive for compatibility


jeromecombes:
merged on master

PR #942, opened 2023-08-31
MT41894: Fix warning undefined array key VEVENT when importing ics calendars

jeromecombes:
MT41894: Fix warning undefined array key VEVENT when importing ics calendars


veggiematts (approved):

jeromecombes:
Force-push = rebase master

PR #947, opened 2023-09-20
MT41956: Fix weekly averages in agent and service statitics

veggiematts:
Test plan:

 - Without the patch, the per-sites weekly averages in agent and service statistics are wrong (the weekly average for all sites is used)

 - With the patch the per-sites weekly averages in agent and service statistics are correctly displayed


jeromecombes (changes requested):
Les tests ne sont pas concluants, voir ticket MT41956

jeromecombes:
force-push = rebase master

PR #949, opened 2023-09-28
MT42195: Add the ICS refresh-interval property

jeromecombes:
MT42195: Add the ICS refresh-interval property


jeromecombes:
Tests show no progression. I close the PR

PR #956, opened 2023-10-12
MT42170: Fix undefined array key in public/ics/calendar.php (for 22.10 and 23.04)

jeromecombes:
MT42170: Fix undefined array key in public/ics/calendar.php

22.10.xx and 23.04.xx only


veggiematts (approved):

jeromecombes:
Merged on 22.10.xx and 23.04.xx
Already fixed on master and 23.10.xx

PR #958, opened 2023-10-13
MT42400: Add the ICS refresh-interval property (for 22.10 and 23.04)

jeromecombes:
MT42400: Add the ICS refresh-interval property (for 22.10 and 23.04)


jeromecombes:
Tests show no progression. I close the PR

PR #959, opened 2023-10-19
MT41778: CircleCI: Download Google Chrome v114 from archive for compatibility

jeromecombes:
MT41778: CircleCI: Download Google Chrome v114 from archive for compatibility


jeromecombes:
merged on 22.10.xx and 23.04.xx
already on master

PR #966, opened 2023-11-03
MT40009: Fix comp time display in 'My Account' when negative

veggiematts:
Test plan:
 - Have a negative comp time credit of half an hour
 - Without the patch, the display in 'My Account' shows -1h30
 - With the patch, it correctly shows -0h30

Unit tests: tests/Include/IncludeFunctionTest.php


jeromecombes (approved):
revue et tests : OK
PR approuvée pour master, 23.04.xx et 23.10.xx

PR #978, opened 2023-12-01
MT42815: Escape teleworking reasons when sending planning notifications

jeromecombes:
MT42815: Escape teleworking reasons when sending planning notifications


veggiematts (approved):

jeromecombes:
Merged

PR #980, opened 2023-12-12
MT42875: Week number in week cycle should not be reset on new year

veggiematts:
For configuration option nb_semaine >= 5, week cycles are not reset on new year anymore.

When applying the patch, dateDebutPlHebdo is updated to keep a continuity in week cycles between the previous behavior and the current one.

Unit tests: tests/Class/ClassDatePlTest.php


jeromecombes (approved):
tests et revue : OK
approuvée pour 23.04.xx, 23.10.xx et master

PR #983, opened 2023-12-22
MT43036: Fix a calculation error on statistics per agent

jeromecombes:
MT43036: Fix a calculation error on statistics per agent


jeromecombes:
Merged

PR #984, opened 2023-12-28
MT43082: src/Cron commands return 0 on success and 1 on error

jeromecombes:
MT43082: src/Cron commands return 0 on success and 1 on error


veggiematts (approved):

jeromecombes:
Merged

PR #987, opened 2024-01-16
MT43250: Import MS Graph : ignore statuses free and tentative

jeromecombes:
MT43250: Import MS Graph : ignore statuses free and tentative


jeromecombes:
Merged

PR #988, opened 2024-01-23
MT42522: ICS Import: import the description depending on the configuration

jeromecombes:
MT42522: ICS Import: import the description depending on the configuration


jeromecombes:
Merged

PR #989, opened 2024-01-25
MT43165: deny overtime change when validated level 1

jeromecombes:
MT43165: deny overtime change when validated level 1


jeromecombes:
Merged

PR #992, opened 2024-01-26
MT42886: fix the calculation of the forecast credit of compensatory time

jeromecombes:
MT42886: fix the calculation of the forecast credit of compensatory time


jeromecombes:
Merged

PR #994, opened 2024-01-30
MT35818: Remove locks from plannings before imports

jeromecombes:
MT35818: Remove locks from plannings before imports

### DIFF
--- a/src/Controller/IndexController.php
+++ b/src/Controller/IndexController.php
@@ -546,6 +546,17 @@ class IndexController extends BaseController
             $dates[0] = $date;
         }
 
+        // Remove locks from associated plannings
+        $start = $dates[0];
+        $end = end($dates);
+
+        $db = new \db();
+        $db->CSRFToken = $CSRFToken;
+        $db->delete('pl_poste_verrou', array(
+            'site' => $site,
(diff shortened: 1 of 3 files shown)


src/Controller/IndexController.php
```diff
@@ -546,6 +546,17 @@ public function model_import(Request $request, Session $session)
             $dates[0] = $date;
         }
 
+        // Remove locks from associated plannings
```
veggiematts, line 549: Technically correct, but for new code like this, shouldn't we be doing it the Symfony way?

see src/PlanningBiblio/DataPurger.php for PlanningPositionLock deletion based on dates (simplePurge function)

I would recommand creating an src/Repository/PlanningPositionLockRepository.php and add a function there to take care of the deletion.

veggiematts (changes requested):

src/Controller/IndexController.php
jeromecombes: done: 827973e57c7fe3a5b3c6007b47e27883bf101161

veggiematts (approved):

jeromecombes:
Merged

PR #999, opened 2024-02-05
MT40653: Right copy does not return to top of page

jeromecombes:
MT40653: Right copy does not return to top of page

### DIFF
--- a/public/planning/poste/fonctions.php
+++ b/public/planning/poste/fonctions.php
@@ -152,7 +152,7 @@ function cellule_poste($date, $debut, $fin, $colspan, $output, $poste, $site)
         $cellule.="<div id='cellule{$GLOBALS['idCellule']}_$i' class='cellDiv {$classe[$i]} pl-cellule-perso-{$resultats[$i]['perso_id']}' {$color[$i]} data-perso-id='{$resultats[$i]['perso_id']}'>{$resultats[$i]['text']}</div>";
     }
 
-    $cellule .= '<a class="pl-icon arrow-right"></a>';
+    $cellule .= '<a class="pl-icon arrow-right" role="button"></a>';
     $cellule.="</td>\n";
     return $cellule;
 }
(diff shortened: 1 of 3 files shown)


public/planning/poste/fonctions.php
```diff
@@ -152,7 +152,7 @@ function cellule_poste($date, $debut, $fin, $colspan, $output, $poste, $site)
         $cellule.="<div id='cellule{$GLOBALS['idCellule']}_$i' class='cellDiv {$classe[$i]} pl-cellule-perso-{$resultats[$i]['perso_id']}' {$color[$i]} data-perso-id='{$resultats[$i]['perso_id']}'>{$resultats[$i]['text']}</div>";
     }
 
-    $cellule .= '<a class="pl-icon arrow-right" href="#"></a>';
+    $cellule .= '<a class="pl-icon arrow-right"></a>';
```
veggiematts, line 155: I would perhaps add role="button" for accessibility:
`$cellule .= '<a class="pl-icon arrow-right" role="button"></a>';`
See https://developer.mozilla.org/fr/docs/Web/Accessibility/ARIA/Roles/button_role

public/planning/poste/js/planning.js
```diff
@@ -1440,7 +1440,7 @@ function bataille_navale(poste,date,debut,fin,perso_id,barrer,ajouter,site,tout,
 
       // Ajout du widget pour copier les agents dans
       // la cellule immédiatement à droite.
-      $("#td"+cellule).append('<a class="pl-icon arrow-right" href="#"></a>');
+      $("#td"+cellule).append('<a class="pl-icon arrow-right"></a>');
```
veggiematts, line 1443: I would perhaps add role="button" for accessibility:
`$("#td"+cellule).append('<a class="pl-icon arrow-right" role="button"></a>');`
See https://developer.mozilla.org/fr/docs/Web/Accessibility/ARIA/Roles/button_role

veggiematts (changes requested):

veggiematts (approved):

jeromecombes:
Merged

PR #1000, opened 2024-02-05
MT43319: allow the deletion of non-validated holiday

jeromecombes:
MT43319: allow the deletion of non-validated holiday


veggiematts (approved):

jeromecombes:
Merged

PR #1001, opened 2024-02-05
MT43489: Allow user to delete their own absences with the correct ACL

jeromecombes:
MT43489: Allow user to delete their own absences with the correct ACL


veggiematts (approved):

jeromecombes:
Merged

PR #1004, opened 2024-02-19
MT43534: Fix PresentSet::all when called in loop

jeromecombes:
MT43534: Fix PresentSet::all when called in loop


veggiematts (approved):

jeromecombes:
Merged master, 23.04.xx and 23.10.xx

PR #1006, opened 2024-02-27
MT42628: Purge calendars when their checkbox is unchecked

jeromecombes:
MT42628: Purge calendars when their checkbox is unchecked


veggiematts (approved):

jeromecombes:

PR #1017, opened 2024-03-08
MT43804: Optimize PresentSet::all

jeromecombes:
MT43804: Optimize PresentSet::all


jeromecombes:

PR #1020, opened 2024-03-15
MT43887: update database via composer

jeromecombes:
MT43887: update database via composer


jeromecombes:
Merged on master and 24.04.xx

PR #1022, opened 2024-03-20
MT43926: Test if the URL is valid in CJICS::purge

jeromecombes:
MT43926: Test if the URL is valid in CJICS::purge


veggiematts (approved):

jeromecombes:
Merged on master, 23.04.xx and 23.10.xx

PR #1023, opened 2024-03-22
MT43653: Count holidays in statistics. Fix using wrong variable

jeromecombes:
MT43653: Count holidays in statistics. Fix using wrong variable


jeromecombes:
Merged on master

PR #1024, opened 2024-03-28
MT43984: Display the correct date on non-validated schedules in week view

jeromecombes:
MT43984: Display the correct date on non-validated schedules in week view


btangaro:
Functional tests OK

jeromecombes:
Merge on master, 23.04.xx and 23.10.xx

PR #1026, opened 2024-04-10
MT44098: Show used positions only in statistics

jeromecombes:
MT44098: Show used positions only in statistics


jeromecombes:
PR 1026 mergée sur master (pour la future version 24.10)

PR #1029, opened 2024-04-11
MT44129: HolidayRepository::get no longer return deletion information

jeromecombes:
MT44129: HolidayRepository::get no longer return deletion information


jeromecombes:
PR 1029 mergée sur master, 23.10.xx et 24.04.xx

PR #1034, opened 2024-04-22
MT44224: Fixes information about events imported from Office 365

jeromecombes:
MT44224: Fixes information about events imported from Office 365


jeromecombes:
PR 1034 mergée sur master, 23.10.xx et 24.04.xx

PR #1039, opened 2024-05-02
MT44329: When multiple recipents are given, Mail-From is used as TO

veggiematts:
When multiple recipients are given to CJMail, they are all set to BCC for PHPMailer.

This led to TO being empty, which we want to avoid.

This patchs adds system configuration Mail-From as TO when all recipients are set to BCC.

Unit tests: tests/PlanningBiblio/NotifierTest.php

### DIFF
--- a/public/include/function.php
+++ b/public/include/function.php
@@ -257,13 +257,8 @@ class CJMail implements NotificationTransporterInterface
         $this->message = $message;
     }
 
-  
-    public function send()
+    public function setPHPMailer()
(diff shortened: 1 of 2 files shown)


jeromecombes (approved):
revue et tests OK
approuvée pour master, 24.04 et 23.10

jeromecombes:
PR 1039 mergée sur master, 23.10.xx et 24.04.xx

PR #1041, opened 2024-05-13
MT43652: Login case insensitive with OpenIDConnect

jeromecombes:
MT43652: Login case insensitive with OpenIDConnect


jeromecombes:
PR 1041 mergée sur master, 23.10.xx et 24.04.xx

PR #1042, opened 2024-05-17
MT44262: Compatibility PHP 8.3, remove depreciations, fix unit tests

jeromecombes:
MT44262: Compatibility PHP 8.3, remove depreciations, fix unit tests


jeromecombes:
PR 1042 mergée sur master (pour la future version 24.10)

PR #1046, opened 2024-06-05
MT44586: Remove HTML entities from framework titles

jeromecombes:
MT44586: Remove HTML entities from framework titles


jeromecombes:
replaced with #1054

jeromecombes:

PR #1053, opened 2024-07-19
MT44663: Planook: Disable cell graying

jeromecombes:
MT44663: Planook: Disable cell graying


jeromecombes:

PR #1054, opened 2024-07-19
MT45026: Planook: Allow sending passwords by email

jeromecombes:
MT45026: Planook: Allow sending passwords by email


jeromecombes:

PR #1069, opened 2024-08-28
MT44970: Remove alert interval greater than one year on absence search

jeromecombes:
MT44970: Remove alert interval greater than one year on absence search


jeromecombes:
Correctif intégré aux versions 23.10.15, 24.04.04 et master (24.05.04)

PR #1075, opened 2024-09-23
MT45495: Allow edition of absences that include deleted agents

jeromecombes:
MT45495: Allow edition of absences that include deleted agents


jeromecombes:

PR #1076, opened 2024-09-24
MT45462: fix initialization of drop-down menus of the manager module

jeromecombes:
MT45462: fix initialization of drop-down menus of the manager module


jeromecombes:

PR #1077, opened 2024-09-26
MT45641: Send a separate email to each recipient (no longer use Cci)

jeromecombes:
MT45641: Send a separate email to each recipient (do not use Cci)


jeromecombes:

PR #1081, opened 2024-10-09
MT45847: fix recording of negative recovery credits

jeromecombes:
MT45847: fix recording of negative recovery credits


jeromecombes:
Correctif intégré aux versions 23.10.18, 24.04.07, 24.10.00 et master (24.05.08)

PR #1083, opened 2024-10-28
Issue #394: Check if a holiday exists before deleting it

jeromecombes:
This prevents recrediting balances several times.
Also improves the sorting of dates on holiday list.


jeromecombes:

PR #1088, opened 2024-11-21
MT44566: Adapt OpenID connector to Google

jeromecombes:
MT44566: Adapt OpenID connector to Google

Also redirect to the login page after logout when OpenID is not used


veggiematts (approved):

jeromecombes:

PR #1089, opened 2024-11-22
MT44985: Framework hours edition: Next start time change only if the …

jeromecombes:
…previous end is modified

Without this patch, when we change an start time, the next start time is changed. With this patch, the next start time change only if the end is modified (expected behavior).


veggiematts (approved):

jeromecombes:

PR #1091, opened 2024-12-06
MT46579: Agents are no longer displayed "deleted"

jeromecombes:
Agents are no longer displayed "deleted" in schedule notifications if absences are ignored according to the Absences-Exclusion parameter


jeromecombes:
Correctif intégré à master (24.11.03) et 24.10.xx (24.10.03)

PR #1093, opened 2024-12-06
MT46575: ICS Export: Schedules are exported if absences are ignored

jeromecombes:
Schedules are exported if absences are ignored according to the Absences-Exclusion parameter


jeromecombes:
Correctif intégré à master (24.11.03) et 24.10.xx (24.10.03)

PR #1098, opened 2024-12-13
MT46682: do not cross out positions compatible with teleworking when …

jeromecombes:
MT46682: do not cross out positions compatible with teleworking when there are teleworking absences

in calendars and ICS export


jeromecombes:
Correctif intégré à master, 24.10.xx et 24.04.xx

PR #1100, opened 2024-12-17
MT45706: Use a helper to convert durations from decimal to hh:mm and vice-versa

veggiematts:
Such conversions were spread across the code and could cause inbetween minutes values stored in the database, as well as errors and discrepancies in displaying converted values between different places in the application.

Two function have been added to the HourHelper class:

    - decimalToHoursMinutes($decimal_duration)
    - hoursMinutesToDecimal($hours, $minutes)

decimalToHoursMinutes takes care of possible incorrect values (decimal value inbetween two minutes) by giving the closest minute.

More importantly, decimalToHoursMinutes and hoursMinutesToDecimal are bijective (after rounding to the closest minute): going back and forth from decimal to hh:mm and vice-versa will always give the same results.
The testHoursMinutesToDecimalAndDecimalToHoursMinutesBijection makes sure of that.

The agent profile now uses HourHelper to display and store comp times.

Test plan:

 1) Go to an agent profile and edit the comp time

 2) Check that it has been properly stored and can be displayed again.

 3) You can fiddle with the comp_time value in the personnel table in the
    database and insert incorrect inbetween minutes values to check that the
    closest minute is always selected. You can use tricky values like inbetween
    59 minutes and the next hour. See HourHelperTest.php for examples.

Next step, when this has been proven to work well: use it everywhere such conversion is made.

### DIFF
--- a/public/conges/class.conges.php
+++ b/public/conges/class.conges.php
@@ -27,6 +27,7 @@ require_once __DIR__."/../absences/class.absences.php";
 use App\PlanningBiblio\WorkingHours;
 use App\PlanningBiblio\ClosingDay;
 use App\PlanningBiblio\Helper\HolidayHelper;
+use App\PlanningBiblio\Helper\HourHelper;
 use App\Model\Agent;
 
 class conges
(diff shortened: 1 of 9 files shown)


jeromecombes (approved):

jeromecombes:
Correctif PR 1100 intégré à master, 24.04.13 et 24.10.07.

PR #1101, opened 2024-12-18
MT46717: Enhance Statistics by Position

jeromecombes:
Enhance the query used to get used positions by adding "group by"
There are now fewer results to analyze in the loops and the response time is better


veggiematts (approved):

jeromecombes:
intégré à master et 24.10.xx

PR #1107, opened 2025-01-27
MT46949: Fix getRecipients to handle empty values for Absences-notifications-*

veggiematts:
In planook, configuration values for Absences-notifications-* are empty strings (and not empty arrays in JSON)

This patch fixes the getRecipients function so it handles empty values for Absences-notifications-*

Test plan:

 1) disable Absences-notifications-agent-par-agent and set Absences-notifications-* to empty values:
 update config set valeur='' where nom='Absences-notifications-A1' limit 1;
 update config set valeur='' where nom='Absences-notifications-A2' limit 1;
 update config set valeur='' where nom='Absences-notifications-A3' limit 1;
 update config set valeur='' where nom='Absences-notifications-A4' limit 1;
 update config set valeur='' where nom='Absences-notifications-B1' limit 1;
 update config set valeur='' where nom='Absences-notifications-B2' limit 1;
 update config set valeur='' where nom='Absences-notifications-B3' limit 1;
 update config set valeur='' where nom='Absences-notifications-B4' limit 1;

 2) Try to add or delete an absence and check that you have an error

 3) Apply the patch

 4) Try to add or delete an absence and check that the error is gone

### DIFF
--- a/public/absences/class.absences.php
+++ b/public/absences/class.absences.php
@@ -1161,6 +1161,9 @@ class absences
 
         $categories=$GLOBALS['config']["{$type}-notifications{$validation}"];
         $categories=json_decode(html_entity_decode(stripslashes($categories), ENT_QUOTES|ENT_IGNORE, 'UTF-8'), true);
+        if (!is_array($categories)) {
+            $categories = array();
+        }
         /*
         $categories : Catégories de personnes à qui les notifications doivent être envoyées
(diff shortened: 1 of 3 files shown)


jeromecombes (approved):
revue et tests OK
approuvée

jeromecombes:
Intégrée à master, 24.04 et 24.10

PR #1108, opened 2025-01-27
MT45507: Add position parameter to include it in public service quota

veggiematts:
- A new parameter is added to the position page "take hours into account for public service quota"

 - The bootstrap for unit tests now processes the atomicupdates before running the tests

Test plan:

 - On a planning, add an hour with a position that have the new parameter enabled
 - Check that the time statistics (/statistics/time) take this new hour into account
 - On a planning, add an hour with position that doesn't have the new parameter enabled
 - Check that the time statistics don't take this new hour into account

### DIFF
--- a/public/planning/poste/class.planning.php
+++ b/public/planning/poste/class.planning.php
@@ -112,7 +112,7 @@ class planning
     }
 
     // Affiche la liste des agents dans le menudiv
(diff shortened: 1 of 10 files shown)


jeromecombes (commented):
Le modèle "Position" doit pouvoir retourner "array()" par défaut pour l'attribut "categories".
Ce qui permettrait de simplifier son utilisation dans les contrôleurs : 

```
- $categories = $position->categories() ? $position->categories() : array();
+ $categories = $position->categories();

Je teste cette PR dès que possible
```

jeromecombes:
Intégrée à master et 24.10

PR #1111, opened 2025-02-07
Mt47194: Allow users to delete their own absences

jeromecombes:


jeromecombes:
Intégrée à master et 24.10

PR #1122, opened 2025-03-07
MT47502: Apply Absences-Exclusion to statistics: agent/service/status

veggiematts:
MT46229: Allow the assignment of absent agents to schedules introduced the new Absences-Exclusion parameter.

However, it was only applied to statistics/absence.

This patch applies it to:

 - statistics/agent
 - statistics/service
 - statistics/status

through the common function of the StatisticsController


jeromecombes (changes requested):
Voir https://suivi.biblibre.com/view.php?id=47502#c279476

jeromecombes (approved):
revue et tests : OK
approuvée pour master et 24.10

PR #1135, opened 2025-04-18
MT48103: Add a setting to ignore lunch breaks

jeromecombes:
- Add a new setting named Planning-IgnoreBreaks (configuration / Administration / Planning)
- Boolean, disable by default
- When enable, breaks defined in working hours are ignored when adding agents to the schedule or when importing a model


jeromecombes:
PR 1135 intégrée à master et 25.04.00

PR #1136, opened 2025-04-24
MT46909: Move favicon.ico to the root of the public folder

jeromecombes:
- This avoid redirection error when antispams try to get /favicon.ico
- Also rename logo-favicon to logo-thumbnail
- Also delete unused file include/header.php

TEST PLAN :
- create a HTTP POST request to /unsubscribe/whatever
- Check access log
- Without this patch, a GET request to /favicon.ico is generated and returns 302
- With this patch, the GET request to /favicon returns 200


jeromecombes:
Intégré sur master, 24.10 et 25.04

PR #1137, opened 2025-04-28
MT46909: Add a new setting to disable the unsubscribe link in the email header

jeromecombes:
MT46909: Add a new setting to disable the unsubscribe link in the email header

- The new setting is Mail-UnsubscribeLink, boolean, enabled by default.
- When enable, unsubscribe links are set in email header. When disable, links are not present.
- This patch also add an order for email settings and change some comments, especially the remplacement of HTML entities


jeromecombes:
Intégré sur master, 24.10 et 25.04

PR #1143, opened 2025-06-12
MT48793: Improve missing category A display

veggiematts:
The "missing category A" warning is currently displayed if needed every time a cell is updated in the planning, and removed after a timeout.

This can lead to multiple identical warnings displayed at the same time.

This patch displays the warning only once, and removes it after a cell update (and not based on a timeout)

Test plan:

 1) Edit a planning without category A agent at the end of service
 2) Check that the warning is displayed
 3) Make an edit to the planning, still without category A agent
    at the end of service
 4) Check that the warning is not displayed twice
 5) Add a category A agent at the end of service
 6) Check that the warning is removed


jeromecombes:
PR 1143 intégrée à master, 24.10 et 25.04

PR #1145, opened 2025-06-20
MT49108: we can change our password if we use local authentication

jeromecombes:
the admin account can change its password regardless of the authentication method
other accounts can change their password if they use the local auth method.

### DIFF
--- a/src/Controller/AccountController.php
+++ b/src/Controller/AccountController.php
@@ -21,12 +21,6 @@ class AccountController extends BaseController
     public function index(Request $request, Session $session)
     {
         // Initialisation des variables
-        // Working hours
-        // Années universitaires (si utilisation des périodes définies)
-        $tmp = array();
-        $tmp[0] = date("n") < 9 ? (date("Y")-1)."-".(date("Y")) : (date("Y"))."-".(date("Y")+1);
-        $tmp[1] = date("n") < 9 ? (date("Y"))."-".(date("Y")+1) : (date("Y")+1)."-".(date("Y")+2);
-        $message = null;
         $CSRFSession = $GLOBALS['CSRFSession'];
         $credits = array();
         $perso_id = $session->get('loginId');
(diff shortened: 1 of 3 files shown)


veggiematts (changes requested):
Here are the results I have with the patch:

Can the user change its own password?

Can change password?

SQL:
 admin: 1
 other: 1

LDAP:
 admin: 1
 other: 0

LDAP-SQL:
 admin: 1
 other: 1

CAS:
 admin: 1
 other: 1

CAS-SQL:
 admin: 1
 other: 1

OpenIDConnect:
 admin: 1
 other: 1

Based on the patch commit message:
the admin account can change its password regardless of the authentication method : OK
other accounts can change their password if they use the local auth method: KO

=> Non-admin users can change their password when Auth-Mode is set to CAS or OpenIDConnect. They shouldn't be able to.

What about Auth-Mode=*-SQL ? Should we let non-admin users change their password ? My feeling is that we should, so this is ok.

jeromecombes:
> Here are the results I have with the patch:
> 
> Can the user change its own password?
> 
> SQL: admin: 1 other: 1
> 
> LDAP: admin: 1 other: 0
> 
> LDAP-SQL: admin: 1 other: 1
> 
> CAS: admin: 1 other: 1
> 
> CAS-SQL: admin: 1 other: 1
> 
> OpenIDConnect: admin: 1 other: 1
> 
> Based on the patch commit message: the admin account can change its password regardless of the authentication method : OK other accounts can change their password if they use the local auth method: KO
> 
> => Non-admin users can change their password when Auth-Mode is set to CAS or OpenIDConnect. They shouldn't be able to.
> 
> What about Auth-Mode=*-SQL ? Should we let non-admin users change their password ? My feeling is that we should, so this is ok.

@veggiematts
Le premier contrôle qui désactive la modification du mot de passe se fait avec `$_SESSION['oups']['Auth-Mode'] == 'CAS'` (info stockée en session lors de l'authentification, pas le param stocké en config)
Quand on s'authentifie en SSO avec CAS ou OpenID, `$_SESSION['oups']['Auth-Mode']` est égale à 'CAS'.
Il n'est pas égale à CAS quand on shunte la redirection vers le SSO avec login?noCAS dans l'URL
Le contrôle de cette session permet les modifications des mots de passe de tous lorsqu'ils shuntent en passant par login?noCAS avec les config -SQL.
Pour le reste, dès qu'ils passe en SSO CAS ou OpenID, ce contrôle empêche la modification du mot de passe.

Selon moi : 
SQL:
admin: 1
other: 1

LDAP:
admin: 1
other: 0

LDAP-SQL:
admin: 1
other: 1

CAS:
admin: 1
other: 0 (0 car `$_SESSION['oups']['Auth-Mode'] == 'CAS'` )

CAS-SQL:
admin: 1
other: 0/1 (0 quand `$_SESSION['oups']['Auth-Mode'] == 'CAS'` , 1 quand ils ajoutent login?noCAS dans l'URL)

OpenIDConnect:
admin: 1
other:  0/1 (0 quand `$_SESSION['oups']['Auth-Mode'] == 'CAS'` , 1 quand ils ajoutent login?noCAS dans l'URL)

jeromecombes:
correctif intégré à master, 24.10.14 et 25.04.02

PR #1166, opened 2025-09-16
MT48762: Remove the background color from printing

Xren200:
-Changed the "highlight current user" logic from using inline background color to adding the current-user-cell CSS class in the element and class list, so that the highlight is handled through frontend styling. -In public/planning/poste/ajax.updateCell.php, add a boolean flag field is_current_user to the $tab[$i] array and pass it to public/planning/poste/js/planning.js. If it is the current user, append ' current-user-cell' to classes.

TEST PLAN
-Create and validate a planning, then use a standard user accompte -Use data10 (terminal : mariadb planno < ~/planno_tools/data/data10.sql && composer install) -Load a model,lock the planning,add all the "plannings" rights of the "médiathèque" site to Aurelie -Logout and login with the following credentials : aurelie.aurelie / password -When you are logged in with this account, you can see Aurélie cells in yellow, that's OK. -Delete Aurelie from one cell and then add Aurelie again,the yellow highlight show still. -Configure Firefox to get background colors printed (ctrl + P, Options, check "Imprimer les arrières-plans". Do it only once.) -On the preview, check if Aurélie cells are in yellow(no yellow is correct)

### DIFF
--- a/public/planning/poste/ajax.updateCell.php
+++ b/public/planning/poste/ajax.updateCell.php
@@ -327,9 +327,11 @@ for ($i=0;$i<count($tab);$i++) {
     $tab[$i]["service"]=removeAccents($tab[$i]["service"]);
 
     // Color the logged in agent.
-    $tab[$i]['color'] = null;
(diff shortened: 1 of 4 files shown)


jeromecombes:
Replaced with #1167

PR #1170, opened 2025-09-18
MT49909: use Doctrine Migration

jeromecombes:
Move the column type change (ENUM to TINYINT) to legacy/update.php To avoid the "Data truncated for column 'supprime'" error during migrations


jeromecombes:
merged on master (25.05.08)

PR #1234, opened 2025-11-24
MT50832: holiday_reset_credits_command- use entitymanager

Xren200:


src/Command/HolidayResetCreditsCommand.php
```diff
@@ -81,24 +81,35 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         // Modifie les crédits
-        $db = new \db();
-        $db->CSRFToken = $CSRFToken;
+        $personnel = $this->entityManager->getRepository(Agent::class)->findAll();
```
jeromecombes: Use $agents and $agent instead of $personnel and $p

jeromecombes (changes requested):

PR #1236, opened 2025-11-24
MT50832: holiday_reset_remainder- use entitymanager

Xren200:


jeromecombes (changes requested):
There is already a PR for theses changes : #1235

PR #1238, opened 2025-11-24
MT50832: workinghour_import_command- use entitymanager

Xren200:


src/Entity/WorkingHour.php
```diff
@@ -94,4 +94,33 @@ public function setUser(?array $user): static
...
+    }
+
+    public function setKey(?string $key): static
+    {
+        $this->cle = $key;
+
+        return $this;
+    }
+
+    public function setAll(?array $insert): static
```
jeromecombes: Remove this function, use officials setters in WorkingHourImportCommand

jeromecombes (changes requested):

PR #1354, opened 2026-04-03
MT46629: Use bootstrap for position's modals

lau-fst:
Both group and floor modals.


jeromecombes (approved):

jeromecombes:
PR intégrée à master, release v25.11.22

PR #1358, opened 2026-04-08
MT52853: Resolve broken display dataTable select

lau-fst:
- The display on chrome and vivaldi is fixed
- The display on firefox stays the same ( white on black bg). Custom css doesn't work but text is still readable.
  -  Apparently this is a known issue with firefox browser, where option tag can't be styled through CSS.


jeromecombes (approved):

jeromecombes:
Correctif intégré à master et 25.10.xx

PR #1359, opened 2026-04-09
MT46629 : Use bootstrap for delete agent modals

lau-fst:
- BS Modal for delete and delete validation
- replaced the deleteAgent function with POST method on agent.delete path
- modified the Controller accordingly


jeromecombes (approved):

jeromecombes:
intégré à master

PR #1375, opened 2026-05-06
MT46629: Replace the update login modal with an inline field

lau-fst:
- deleted update-login modal
 - new inline input for the login
 - same comportment as before but with invalid messages
 - translation enabled on said messages
 - used BS for agent edit tabs
 - disabled tab link when input is invalid (only cancel button is enabled)
 - renamed class .modal-tabs to .container-tabs
 - updated tests


jeromecombes (approved):

jeromecombes:
Merged into master (181ae19f19fbf241f578665ecea1679511f2c6be)

PR #1382, opened 2026-05-11
MT52706: Lunch positions hours are not counted in SP hours

jeromecombes:
* Same behavior on schedules and statistics by time.
* The current cell is no longer counted on schedules.


jajm (approved):
Testé OK

jeromecombes:
Mergée sur 25.04, 25.10 et master

PR #1398, opened 2026-06-02
MT46629: Use Bootstrap alerts in group edition

lau-fst:
- BS grid + validity comportment on group edition form
- required attribute on necessary form fields
- removed onsubmit ctrl, the submission is blocked until every necessary field is valid 
- model-name duplicate check with BS on keyup  (replace ctrl_nom fct)
- text_validation fct to sanitize and verify the validity of a text input
- enabled message display when required text input is invalid
- enabled inline-feedback on .form-control elements with a MutationObserver object 
 
TEST PLAN :
 
 - try to submit the form empty
 - verify that the invalidity icon disappear when fields are filled
 - enter a model name that already exists and verify that the invalidity message changes accordingly

### DIFF
--- a/public/js/framework.js
+++ b/public/js/framework.js
@@ -155,28 +155,6 @@ function supprime_tab(nom,id){
 }
 //    --------------------------------    FIN Tableaux - Lignes    -------------------------    //
 
-function ctrl_nom(me){
-  exist=false;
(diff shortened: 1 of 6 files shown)


jeromecombes (approved):

jeromecombes:
Replaced with #1415 (rebased)

PR #1401, opened 2026-06-08
MT53517: Use CJDataTable class for absences CSV import

veggiematts:
Use CJDataTable class for absences CSV import instead of using custom DataTable code:

 - allow to hide column using dataTableHide
 - add log level filtering

Both features can be re-used for other DataTables.

Test plan:
 DataTable behavior should be the same as before, see
 MT48492: Import absences from CSV file:

 Once the file is processed check that you have a table with the results.
 Traditional log levels are used: ERROR, WARNING, INFO, DEBUG
 INFO is used by default, but you can set the log level you want using the
 dropdown list.

### DIFF
--- a/public/js/CJScript.js
+++ b/public/js/CJScript.js
@@ -8,6 +8,24 @@ Fichier : CJScript.js
 
  DataTable.type('date', 'className', 'dt-left');
 
+function CJDataTableLogLevelFilter() {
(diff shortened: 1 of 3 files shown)


jeromecombes (approved):

jeromecombes:
Merged into master and 26.04.xx

PR #1408, opened 2026-06-15
BugFix : duplicate name in select on edit absence

lau-fst:
Capture d’écran_2026-06-15_09-51-56

In edit mode, it was possible to select the same name twice from the list of agents affected by the absence.
This issue does not create a duplicate in the database; it only affects the display.


jeromecombes (approved):

jeromecombes:
Merged into master (26.05.02), 26.04.xx (26.04.02) and 25.10.xx (25.10.21)

PR #1414, opened 2026-06-18
MT53893: Show only undeleted skills in agent records

jeromecombes:
All skills were displayed in agent profiles and in the group edit window.
This fix filters out deleted activities.

TEST PLAN
* Add skills
* Delete some of them
* Edit an agent profile / go to the skills tab (activités)
* Deleted skills should not be displayed

* Select agents from the list
* Make a bulk edit
* Go to the skills tab (activités)
* Deleted skills should not be displayed


jeromecombes:

PR #1415, opened 2026-06-19
MT46629: Use Bootstrap alert in group edition

jeromecombes:
* BS grid + validity comportment on group edition form
* Required attribute on necessary form fields
* Removed onsubmit ctrl, the submission is blocked until every necessary field is valid
* Model-name duplicate check with BS on keyup (replace ctrl_nom fct)
* Text_validation fct to sanitize and verify the validity of a text input
* Enabled message display when required text input is invalid
* Enabled inline-feedback on .form-control elements with a MutationObserver object

TEST PLAN :
* Try to submit the form empty
* Verify that the invalidity icon disappear when fields are filled
* Enter a model name that already exists and verify that the invalidity message changes accordingly

### DIFF
--- a/public/js/framework.js
+++ b/public/js/framework.js
@@ -155,28 +155,6 @@ function supprime_tab(nom,id){
 }
 //    --------------------------------    FIN Tableaux - Lignes    -------------------------    //
 
-function ctrl_nom(me){
-  exist=false;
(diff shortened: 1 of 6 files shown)


jeromecombes:

PR #1416, opened 2026-06-22
MT53921: Fix template/workinghour/edit.html.twig

jeromecombes:
* Some div tags were not closed, which prevented the exception from being validated

TEST PLAN :
* edit a working hour table, update it, save it, check if modifications are saved
* edit a working hour table, click on the "create an exception" button, save the exception, check if the exception is created


lau-fst (approved):
Everything works great.

jeromecombes:

PR #1422, opened 2026-06-24
BugFix :  Undefined variables error on Agent Controller

lau-fst:
In AgentController.php two variables (`$sitesSelect,$rights_sites`) where only initiliazed when
 `Multisites-nombre > 1`.
So when the condition wasn't true an undefined variable error was raised, blocking the agent edit page.


jeromecombes:
Done in #1427

PR #1426, opened 2026-06-26
MT54010: Fix the migration that add the PlanningHebdo-resetCycles param

jeromecombes:
* Add IGNORE to the INSERT INTO config query


jeromecombes:

PR #1427, opened 2026-06-26
MT53720: Do not display unamed sites

jeromecombes:
* Under the planning menu
* In agents records, general info tab and ACL tab.

This is useful when we want to remove a site: we simply need to delete its name from the configuration, while waiting for better site management.

TEST PLAN :
* Create sites (config)
* Check that they display correctly and are accessible under the planning menu.
* Check that they display correctly in agent records, info general tab and access right tab.
* Remove a site name from the configuration without changing the number of sites.
* Check that the other sites still display and are accessible.


jeromecombes:

PR #1429, opened 2026-07-01
MT46629 : holiday/add TESTS

lau-fst:
Added extensive tests to verify the comportment of the new holiday/add template that merges conges/add and comptime/add in one.


jeromecombes:
Merge into #1440

PR #1444, opened 2026-07-07
Datepicker: if empty, end open on same month as start

lau-fst:


jeromecombes (approved):

jeromecombes:
Merged on master (26.05.05) and 26.04.05

PR #1446, opened 2026-07-07
Replace aRight classes with text-end

lau-fst:
+ delete aLeft classes (not useful to the display)


jeromecombes:
Merged into master (26.05.04).

PR #1449, opened 2026-07-07
BugFix : Remove is-invalid on hidden form element on /absence/add

lau-fst:


jeromecombes:

PR #1450, opened 2026-07-07
MT53738: Fix TypeError in StatisticController::common

jajm:
Error message was:

Uncaught PHP Exception TypeError:
"count(): Argument #1 ($value) must be of type Countable|array, null given" at StatisticController.php line 232

$db->result is null when the query does not return any row


jeromecombes (approved):

jeromecombes:
Merged on master (26.05.05), 26.04.05 and 25.10.23

PR #1454, opened 2026-07-09
MT46629: Clean uncessary custom classes with BS classes

jeromecombes:
* Replace .bold with .fw-bold
* Replace .center with text-center
* Replace .textAlignRight with .text-end
* Replace text-align:left with .text-start
* Replace .nowrap with .text-nowrap
* Replace .noprint with .d-print-none
* Remove .fw-bold on reason options
* Remove text-center on timepicker and add property in default.css
* Remove .text-start on h3
* Use variables for colors in css file


jeromecombes:
Merged on master (26.05.05)

PR #1457, opened 2026-07-16
MT52706: Allow counting lunch positions towards the public service quota

jeromecombes:


jeromecombes:
Merged on 26.04 and master

PR #1462, opened 2026-07-17
MT54305: Allow SQL fallback authentication with OpenIDConnect

jeromecombes:


jeromecombes:
Merged on 26.04 and master

PR #1463, opened 2026-07-21
MT53418: Fix validation level on deleted agents

veggiematts:
Validation level on a deleted agent was not properly set when Absences-notifications-agent-par-agent is enabled

Test plan:

 1) Create an absence for an agent you are a manager of

 2) Check that you can edit that absence

 3) Delete that agent
    (once from the agents list, and then again from the
    deleted agents list. Check that personnel.supprime = 2
    in the database for that agent, but keep in mind that
    its login is now "deleted_<id>")

 4) Without the patch, check that you cannot edit the absence
 5) With the patch, check that you can edit the absence


jeromecombes (approved):
revue et tests OK

jeromecombes:
Merged on v26.05.10 (master), v26.04.08 and v25.10.26.

PR #1471, opened 2026-07-24
MT46629: Mettre à jour les librairies frontend

lau-fst:


jeromecombes:
Merged on master, except jQuery Update (ae6bd3d4744f91f4f363c6cc465d8e8e0eeae6ce) which causes unit tests to fail.

PR #1472, opened 2026-07-27
Allow unit tests to be run on Sundays

jeromecombes:
* Add an option to the app:planning:control command, to define when the test needs to be simulated.

> bin/console app:planning:control -h
>   Description:
>     Check upcoming schedules and sends a report to the planning team
>
>   Usage:
>     app:planning:control [options]
>
>   Options:
>         --not-really      For testing purposes. Does not send the email but displays it.
>         --when=WHEN       For testing purposes. When the test needs to be simulated. E.g.: --when=next\ tuesday

E.g.: `bin/console app:planning:control --when=next\ tuesday --not-really`

### DIFF
--- a/src/Command/PlanningControlCommand.php
+++ b/src/Command/PlanningControlCommand.php
@@ -28,22 +28,27 @@ class PlanningControlCommand extends Command
 {
     use \App\Traits\LoggerTrait;
 
-    private $entityManager;
-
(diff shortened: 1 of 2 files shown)


jeromecombes:
Merged on v26.05.10 (master) and v26.04.08

PR #1473, opened 2026-07-27
MT44200: Send a notification when a holiday is deleted

jeromecombes:
* Replace AJAX request by a proper form submit with CSRF protection
* Rewrite holiday deletion using Doctrine
* Add a Mailer service that generates the mail content using Twig
* Holiday entity return null for invalid dates


jeromecombes:
Merged on v26.05.10 (master) and v26.04.08.

PR #1475, opened 2026-07-27
MT53636: Use ClamAV antivirus for absence documents upload

jeromecombes:
Test plan:
 - Enable ClamAV in your .env file by providing the path to ClamAV socket: CLAMAV_SOCKET=unix:///var/run/clamav/clamd.ctl

 - Use the eicar test file: https://www.eicar.org/download-anti-malware-testfile/ (also available in tests/data/eicar.com)

 - Add an absence and check that you have a flashbag warning and that the file has not been attached

 - Edit an absence and check that you have an alert and that the file has not been attached

 - Add/Edit an absence with a random file and check that the file has been attached

 - Disable ClamAV and check that files can still be attached

Unit tests (ClamAV is not mocked, it is really tested):
   tests/Service/ClamAvScannerTest.php

Follow up:
 - Update composer.lock
 - Rector
 - Allow translation
 - Add more meaningful error messages
 - Do not process uploaded file if ClamAV is configured but not available
 - Use english for translatable strings
 - Update templates (comply with MT46629: Merge absence templates)
 - Fix condition in scan
 - Remove debug logs


jeromecombes:
Merged on master (v26.05.09)

PR #1476, opened 2026-07-28
MT52609: Add date filters to info list pages (admin, absences, holiday)

jeromecombes:
and display the table even if there are no results

Follow up:
* Improve AbsenceInfoRepository
* Use BaseController::initDate
* Set correct times for start and end into AbsenceInfoRepository
* Fix AbsenceInfoRepository (now extends EntityRepository)
* Rework info repositories and controllers
  * Remove AbstractInfoRepository, duplicate findByDateRange instead
  * Fix BaseController::initDate for when input is empty or invalid
  * Do not use `setTime` on dates passed to the repository method as it modifies the original object, use explicit doctrine type instead to discard the time part
* Fix AdminInfoControllerTest
* Add reset route
* Clean templates

### DIFF
--- a/src/Controller/AbsenceInfoController.php
+++ b/src/Controller/AbsenceInfoController.php
@@ -4,6 +4,7 @@ namespace App\Controller;
 
 use App\Controller\BaseController;
 
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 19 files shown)


jeromecombes:
Merged on master (v26.04.09)

PR #1483, opened 2026-08-27
MT54389: HolidayController: Dedicated routes for edit and update

jeromecombes:
1. Create dedicated routes for Holiday/edit (show form) and Holiday/update (form submission processing)
2. Use the same route and the same method for recording new holiday and modifying existing holiday
3. Use the same route for saving holidays and comp-times
4. Restore HolidayReminderTest


jeromecombes:
Merged on master

PR #1485, opened 2026-09-09
MT53998: Prevent adding the same agent twice to the same cell

jajm:
- Prevent adding the same agent twice to the same cell
- Exclude from contextual menu agents that are already in cell
- Fix random test failures in PlanningJobControllerTest


jeromecombes (approved):

jeromecombes:
Merged on master and 26.04.09

PR #1486, opened 2026-09-10
MT54859: Fix WorkingHours import

jeromecombes:


jeromecombes:
Merged on master
The first commit (fix) is merged on 26.04.09

PR #1490, opened 2026-09-17
MT54917: Use random_bytes to generate ICS codes

jeromecombes:
* Enhances security; the old method generated a code that could be reconstructed.


jeromecombes:
Merged into master and 26.04.09

